PR #117, opened 2021-03-25
Storybook show code fix

### LINKED ISSUE #15: Storybook's 'show code' option displays incorrect code
When navigating the docs of a story on [nldsreactstorybook](http://dhreactstorybook.azurewebsites.net/), the 'show code' option does not always show working code.

For instance, in the Appbar docs: 
show-not-working

The actual story for Appbar looks like this:
```ts  
return (
    <AppBar {...args}>
      <Toolbar>
        <IconButton
          edge='start'
          className={classes.menuButton}
          color='inherit'
          aria-label='menu'
        >
          <MenuIcon />
        </IconButton>
        <Typography variant='h6'>News</Typography>
      </Toolbar>
      <Button>I am a button!</Button>
    </AppBar>
  )
```

I would expect at least working code, all of which should be in the nlds-react-components library (WithStyle is Material UI).

- [x] Investigate how to customize the code example without creating an entire .MD file.
- [x] Check all component stories to make sure `Show code` shows working code. Compile a list of all stories that have broken code examples so that follow-up issues can be made.
- [ ] Follow-up issues have been made

ogjtech:
Close #15. 
Adding the following code:
```typescript
docs: {
    source: {
        type: "code"
    }
}
````

To the `export default {...} as Meta;` in each Storybook file shows the component code as they are defined in the file.
For example, the `Box` component now shows:
```typescript
(args: any) => (
  <Box {...args}>
    <p>This is a Box component.</p>
  </Box>
)
```

In the Docs tab.


DanielvanVliet:
I'm not sure if this is the right approach. Although the displayed code does work now, all information on the arguments for each variant is lost. See button for instance, all variants show the same code now.

bddjong:
> I'm not sure if this is the right approach. Although the displayed code does work now, all information on the arguments for each variant is lost. See button for instance, all variants show the same code now.

Agreed. This is going in the right direction, but we should have working code examples for all story templates. Take a look at the NLDS POC for example: http://storybook.nldesignsystem.nl/?path=/docs/components--backlink

joostvanviegen (commented):
I agree with the comments from @bddjong and @DanielvanVliet and won't be giving an explicit appproval for now.

DanielvanVliet:
It turns out Storybook has three options for the source code in 'show code'. ([storybook docs](https://storybook.js.org/docs/react/writing-docs/doc-blocks#source))

1. Dynamic with `type: dynamic` option, which is the default
2. Render the source code of the story as a whole with `type: code`
3. Pass in a custom string to be rendered in the show code.

Stories which require no setup code, such as Button, should generate usable code in the docs with the default `type: 'dynamic'` option:

reasonable


The more complicated stories such as DatePicker require some setup code (e.g. a React.useState) to function properly. 

The dynamic option does not result in usable code in this case, as all setup code is removed:
bad

Using the second option (rendering the storybook source code) for these more complicated stories is better:
better
However this code does not respond to changes in the properties, and is thus the same for each variant in the story.
And more importantly, is still missing imports such as DateFnsUtils and might not work.


We can also supply storybook with our own string that should be shown in the 'show code'.
This may be different for each variant, but is still static and will not respond to changes in the properties on the docs page.
This also means that we have to ensure that the code works and this can add alot of work and maintenance when API changes are made.

Still providing custom examples is probably the best way forward for the more 'complicated' components, if we always want to display usable code.
Any thoughts? @Gemeente-DenHaag/mdh-rysst @mjcarsjens

joostvanviegen:
> 
> 
> It turns out Storybook has three options for the source code in 'show code'. ([storybook docs](https://storybook.js.org/docs/react/writing-docs/doc-blocks#source))
> 
>     1. Dynamic with `type: dynamic` option, which is the default
> 
>     2. Render the source code of the story as a whole with `type: code`
> 
>     3. Pass in a custom string to be rendered in the show code.
> 
> 
> Stories which require no setup code, such as Button, should generate usable code in the docs with the default `type: 'dynamic'` option:
> 
> reasonable
> 
> The more complicated stories such as DatePicker require some setup code (e.g. a React.useState) to function properly.
> 
> The dynamic option does not result in usable code in this case, as all setup code is removed:
> bad
> 
> Using the second option (rendering the storybook source code) for these more complicated stories is better:
> better
> However this code does not respond to changes in the properties, and is thus the same for each variant in the story.
> And more importantly, is still missing imports such as DateFnsUtils and might not work.
> 
> We can also supply storybook with our own string that should be shown in the 'show code'.
> This may be different for each variant, but is still static and will not respond to changes in the properties on the docs page.
> This also means that we have to ensure that the code works and this can add alot of work and maintenance when API changes are made.
> 
> Still providing custom examples is probably the best way forward for the more 'complicated' components, if we always want to display usable code.
> Any thoughts? @Gemeente-DenHaag/mdh-rysst @mjcarsjens

I think we should go with a hybrid between 'dynamic', 'code' and 'custom'. Whenever a story is simple enough to use dynamic we'll do so, for the other ones we should decide on a case by case basis if 'code' or 'custom' is better.

mjcarsjens:
I personally feel like we should try to use `dynamic` where possible. In cases where this does not work, as for example the DatePicker component my personal preference would go out to writing a custom `code` block, as follows:

```
Inline.parameters = {
  docs: {
    source: {
      code:
`
import DateFnsUtils from "@date-io/date-fns";

<PickersUtilsProvider utils={DateFnsUtils}>
  <DatePicker variant="inline" />
</PickersUtilsProvider>
`
    }
  }
}
```

This does bring along some more maintenance but these are a few rare cases from what I can find, and when API changes are made you would have to update the Story file anyway, so updating 3 code examples for the variants of the component doesn't seem like that much extra work. 

I would try to stay away from `type: "code"` as this does not correctly display how to actually get the variant which is shown. 

With regards to components like the `Accordion` and the `AppBar`: There seems to be a problem with the `@material-ui/icons` components in combination with StoryBook, as these all show up as `<[Object object] />`. I would not put too much effort into fixing this, if the problem still persists once we implemented our own icons this is something we can look at, for now I would simply disregard these issues.

ogjtech:
@Gemeente-DenHaag/mdh-rysst @mjcarsjens 
The `Badge ` component is one such example of showing `[object Object]` when the code is displayed in the `dynamic` variant.
To fix this, I have replaced the type of the `args` parameter to `BadgeProps`, in order to show the Badge as it is defined in the code and to show more specifically what props can be supplied. Badge is now defined as follows:

ogjtech:
> In cases where this does not work, for example the DatePicker component, my personal preference would go out to writing a custom code block

@mjcarsjens Even though this does show the component as is, it can only be defined once in a Storybook file, not per defined variant. It is not technically impossible, but it would require us to rewrite the way we structure our Storybook files and resort to an outdated way of defining the Templates that are shown. This would mean we need to rewrite _all_ of the storybook files, which in my personal opinion is more trouble than its worth.

Personally I think it better to, for now, just accept that the code is not shown as ideal and trust on the implementer's ability to read documentation and interpret code.

mjcarsjens:
> 
> 
> > In cases where this does not work, for example the DatePicker component, my personal preference would go out to writing a custom code block
> 
> @mjcarsjens Even though this does show the component as is, it can only be defined once in a Storybook file, not per defined variant. It is not technically impossible, but it would require us to rewrite the way we structure our Storybook files and resort to an outdated way of defining the Templates that are shown. This would mean we need to rewrite _all_ of the storybook files, which in my personal opinion is more trouble than its worth.
> 
> Personally I think it better to, for now, just accept that the code is not shown as ideal and trust on the implementer's ability to read documentation and interpret code.

@Gemeente-DenHaag/mdh-rysst @ogjtech 
Not exactly true though, you can override the default parameters for each variant:

The code:
image

The result:
image

As you can see defining a code template for an individual variant does only change the Show Code feature for that specific variant and it is very much possible to define a own code definition for each variant (see my experimental commit to take a look at the exact code).

I'm not saying your conclusion is wrong in any way, I think this is a decision that should be made with the team as a whole and I'm not trying to veto what we should do. All I'm trying to say is that the option to provide a custom code template for every variant is still possible and can be considered as an option.

ogjtech:
@mjcarsjens I have tried your approach and, while it does work, it shows very weird indentation. I have tried implementing some of the fixes suggested in [this issue](https://github.com/storybookjs/storybook/issues/8078), but none of them seem to work.
This issue has existed since September 2019, but to this day has not seen a fix of any kind.

ogjtech:
@Gemeente-DenHaag/mdh-rysst All components have been individually evaluated on whether they require a custom string for showing code or whether `dynamic` or `code` suffices. You can see the resulting code by checking out the `storybook_show_code_fix` branch and running Storybook.

I did not bother replacing all the `noRefCheck()` functions with blank functions, as I figured most developers seeing these docs can figure out on their own what props like `onClick`, `onClose`. `onChange`, et cetera mean and what they should contain.

bddjong:
Just a reminder that we use `feature/#nr-description` and `fix/#nr-description` for branch names

src/stories/datadisplay/Avatar.stories.tsx
```diff
@@ -1,17 +1,22 @@
-import React from "react";
-import { Meta, Story } from "@storybook/react";
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
 import { Avatar, AvatarProps } from "../../components/Avatar";
 import pkg from "../../components/Avatar/package.json";
```
bddjong, line 4: I think your prettier or eslint config is doing something weird here. Half of the imports use single quotes and half of them use double quotes.

bddjong (changes requested):

src/stories/datadisplay/Avatar.stories.tsx
ogjtech: Probably did not bother to check everywhere, will fix after merge conflicts are resolved.

src/stories/datadisplay/Divider.stories.tsx
```diff
@@ -1,62 +1,163 @@
...
+            <ListItemIcon>
+              <DraftsIcon />
+            </ListItemIcon>
+            <ListItemText primary='I am under the divider.' />
+          </ListItem>
+        </List>
+      `
+    }
+  }
+}
```
DanielvanVliet, line 67:
```suggestion
var code =  '
const [anchorEl, setAnchorEl] = React.useState(null);
const handleClick = (event) => { setAnchorEl(event.currentTarget); };
const handleClose = () => { setAnchorEl(null); };

return (
  <div>
    <Button aria-controls='menu' onClick={handleClick}>Click me!</Button>
    <Menu
      open={Boolean(anchorEl)}
      anchorEl={anchorEl}
      onClose={handleClose}
      id='menu'
    >
      <MenuItem dense={false} onClick={handleClose}>
        <p>Set the &quot;open&quot; property to <code>true</code> to force the menu to open.</p>
      </MenuItem>
      <MenuItem dense={false} onClick={handleClose}>
        <span>Click me!</span>
      </MenuItem>
      <MenuItem dense={false} onClick={handleClose}>
        <span>Or me!</span>
      </MenuItem>
      <MenuItem dense={false} onClick={handleClose}>
        <span>Or me!</span>
      </MenuItem>
    </Menu>
  </div>
);
'

Default.parameters = {
  docs: {
    source: {
      // language=JS
      code
    }
  }
}
```

DanielvanVliet (changes requested):
Currently there are additional spaces inserted in the code examples.
One way to avoid this would be creating the code variables outside the story config.

Can you do this for all the stories?

src/stories/datadisplay/Divider.stories.tsx
DanielvanVliet: Use the ` quotes instead of ', github doesn't like the use of ` in code samples

GewoonMaarten (approved):

DanielvanVliet (approved):

bddjong (approved):

Robbert:
@ogjtech If that multiline string indentation is such an annoying recurring issue, you could consider using this: https://www.npmjs.com/package/ts-dedent

Robbert:
Also: wouldn't it be convenient to have `prettier --check` in the CI/CD pipeline? If someone doesn't have the Prettier format-on-save installed it becomes immediately apparent, and you don't have to _dweil with de crane open_.

PR #124, opened 2021-04-07
Add base styles package

bddjong:
To prevent duplicate styles, a shared base styles package should be available for all components to depend on.

The globals.module.css contains the styles as defined in the [Design Systems identity](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=43%3A338)

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -29,7 +29,7 @@
     "unit-no-unknown": [true],
     "alpha-value-notation": ["percentage"],
     "hue-degree-notation": ["number"],
-    "color-function-notation": ["legacy"],
+    "color-function-notation": ["modern"],
     "length-zero-no-unit": [true, "custom-properties"],
     "font-weight-notation": [
       "numeric",
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -29,7 +29,7 @@
     "unit-no-unknown": [true],
     "alpha-value-notation": ["percentage"],
     "hue-degree-notation": ["number"],
-    "color-function-notation": ["modern"],
+    "color-function-notation": ["legacy"],
     "length-zero-no-unit": [true, "custom-properties"],
     "font-weight-notation": [
       "numeric",
--- a/src/components/BaseStyles/globals.module.css
+++ b/src/components/BaseStyles/globals.module.css
@@ -1,45 +1,45 @@
 :root {
   /* Colors */
-  --denhaag-green-1: hsl(117, 28%, 87%);
-  --denhaag-green-2: hsl(118, 28%, 60%);
-  --denhaag-green-3: hsl(138, 58%, 33%);
-  --denhaag-green-4: hsl(138, 57%, 27%);
-  --denhaag-green-5: hsl(139, 100%, 9%);
+  --denhaag-green-1: hsl(117 28% 87%);
+  --denhaag-green-2: hsl(118 28% 60%);
+  --denhaag-green-3: hsl(138 58% 33%);
+  --denhaag-green-4: hsl(138 57% 27%);
+  --denhaag-green-5: hsl(139 100% 9%);
 
-  --denhaag-blue-1: hsl(208, 76%, 92%);
-  --denhaag-blue-2: hsl(210, 64%, 80%);
-  --denhaag-blue-3: hsl(207, 80%, 35%);
-  --denhaag-blue-4: hsl(207, 87%, 27%);
-  --denhaag-blue-5: hsl(207, 93%, 16%);
+  --denhaag-blue-1: hsl(208 76% 92%);
+  --denhaag-blue-2: hsl(210 64% 80%);
+  --denhaag-blue-3: hsl(207 80% 35%);
+  --denhaag-blue-4: hsl(207 87% 27%);
+  --denhaag-blue-5: hsl(207 93% 16%);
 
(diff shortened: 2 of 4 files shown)


joostvanviegen (approved):
Stylesheet and package.json look good to me. We might want to include an index.ts or index.tsx file to make importing these styles easier.

src/components/BaseStyles/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@gemeente-denhaag/basestyles",
+  "description": "A BaseStyles component",
+  "author": "Municipality of The Hague",
+  "license": "EUPL-1.2",
+  "main": "index.tsx",
```
joostvanviegen, line 6: Shouldn't this also be index.ts?

mjcarsjens:
No comments from me, looks like a nice first setup to work with!

mjcarsjens:
On second thought, one small comment: NLDS (and our designers) prefer the use of HSL values for color as this is a format which is easier to grasp then hexcodes or rgb values (I agree this is a point of discussion but oh well). In figma you can toggle how the color should show and HSL is an option here (well, technically the values displayed by figma are HSLA but the alpha is always 1 so might as well use HSL).

Now in our case the hue values don't align as nicely as they could, but in an ideal scenario all variants of a color all could have the same hue and saturation, and only have a different light value. This can be seen in the brand values which NLDS has been working on with Utrecht: https://github.com/nl-design-system/utrecht/blob/main/src/static/brand.css 

As you can see these are some very intuitive values, where it is easy to see that all variants of a color only have a different light value but are otherwise the same.

Robbert:
Indeed we think `hsl()` color values are better human readable -- especially in the context of lightness-based design tokens -- than hex or rgb.

For a color scale it isn't necessary for all values to have exactly the same H/S values and only different lightnesses, and the lightnesses don't need to be _exactly_ the same as the design token scale.

So when the design token scale is `some-color-20`/`some-color-40`/`some-color-40`/`some-color-80` it would be fine to have the following values:

`--some-color-20: hsl(120, 30, 20%)`
`--some-color-40: hsl(122, 30, 42%)`
`--some-color-60: hsl(121, 32, 61%)`
`--some-color-80: hsl(120, 31, 80%)`

The idea of having the design tokens named after the rounded lightness value, is also to have a certain degree of freedom to change the actual color values without updating any of the other code. The only limitations you have is that the contrast values between the design tokens should remain equally accessible.

joostvanviegen:
> 
> 
> Indeed we think `hsl()` color values are better human readable -- especially in the context of lightness-based design tokens -- than hex or rgb.
> 
> For a color scale it isn't necessary for all values to have exactly the same H/S values and only different lightnesses, and the lightnesses don't need to be _exactly_ the same as the design token scale.
> 
> So when the design token scale is `some-color-20`/`some-color-40`/`some-color-40`/`some-color-80` it would be fine to have the following values:
> 
> `--some-color-20: hsl(120 30% 20%)`
> `--some-color-40: hsl(122 30% 42%)`
> `--some-color-60: hsl(121 32% 61%)`
> `--some-color-80: hsl(120 31% 80%)`
> 
> The idea of having the design tokens named after the rounded lightness value, is also to have a certain degree of freedom to change the actual color values without updating any of the other code. The only limitations you have is that the contrast values between the design tokens should remain equally accessible.

We will discuss renaming the design tokens with the designer.

ogjtech (approved):
Code looks good, although I'm unsure whether we want to supply these files as a separate package and not just include them in the styling files for components.

Robbert:
I would love to see the The Hague brand design tokens available as CSS variables in its own package! Regardless of what web framework or product you use, importing `globals.module.css` in any codebase for The Hague and applying the colors consistently from a single source of truth, is a great way to apply the design system.

.stylelintrc.json
```diff
@@ -29,7 +29,7 @@
     "unit-no-unknown": [true],
     "alpha-value-notation": ["percentage"],
     "hue-degree-notation": ["number"],
-    "color-function-notation": ["modern"],
+    "color-function-notation": ["legacy"],
```
Robbert, line 32: I would be interested to hear if you encountered any practical issues with the `modern` notation for color values, we could take that perspective in consideration for our own linting rules.

bddjong: This was specifically to use the format you provided in your example. I believe @joostvanviegen received linter warnings when using percentages in the HSL notation.

Robbert: Oh you were quick then! I accidentally had used the old notation, and I had updated my example to use the `modern` instead of the `legacy` notation, but apparently not before you copied it and ran into issues. There are some advantages to using the new modern notation, so I would recommend sticking with the modern notation instead of using my old example.

mjcarsjens (approved):

joostvanviegen (approved):

src/components/BaseStyles/globals.module.css
```diff
@@ -0,0 +1,61 @@
+:root {
+  /* Colors */
+  --denhaag-green-1: hsl(117, 28%, 87%);
+  --denhaag-green-2: hsl(118, 28%, 60%);
+  --denhaag-green-3: hsl(138, 58%, 33%);
```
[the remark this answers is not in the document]
Robbert: You can include it in documentation or in a code comment, not sure why it would be useful to make it a separate token too.

Issue #54, opened 2021-02-27
Design: Button


GewoonMaarten:
The button component needs to be styled according to the [Figma](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1%3A656) design.

The button has been largely styled in the [Design Token POC](https://github.com/Gemeente-DenHaag/denhaag-component-library/pull/40).

Animations are out of scope!

Write down instructions/checklist in the wiki

## Acceptance criteria
**Variants implemented**
- [x] Primary action (filled background)
- [x] Secondary action (outline)

**Sizes implemented**
- [x] Large size
- [x] Medium size

**States implemented**
- [x] Normal
- [x] Hover
- [x] Disabled
- [x] Focus

**Icon support**
- [x] Icon can provided
- [x] Icon can be placed on the left side
- [x] Icon can be placed on the right side

**Requirements**
- [x] Width varies depending on the length of the text
- [x] [Alignment](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1%3A656)  requirements are met
- [x] Classes use the BEM naming convention
- [x] Design tokens are identified > proposed > implemented
- [x] Only the `props` and possible values for these `props` which are necessary in our use-case are exposed. All other Material-UI and standard HTML props which are not deemed necessary in our use-case should not be allowed/forwarded to the underlying component.

mjcarsjens:
In consultation with Robbert:
- [ ] No hardcoded margins: move these to design tokens
- [ ] Move filling of design tokens to a `brand.module.css` file

bddjong:
> In consultation with Robbert:
> 
> * [ ]  No hardcoded margins: move these to design tokens
> * [ ]  Move filling of design tokens to a `brand.module.css` file

Created https://github.com/Gemeente-DenHaag/denhaag-component-library/issues/156 to address the second issue

bddjong:
Design tokens in documetnatie van button in storybook

PR #133, opened 2021-04-14
Button design implementation

bddjong:
This pull request implements issue #54 Design Button

## What's new
- Applied button design 
- Added button design tokens to `globals.module.css` (with the `:root` selector)
- Button now supports icons on the left and right side (this uses the underlying Material UI icon support in buttons)
- A few new story templates to display the different button sizes and buttons with icons

## How to test
Run storybook locally (`yarn storybook`) and check out the button page to see the new design.


src/components/Button/button.module.css
```diff
@@ -29,7 +28,9 @@
   --denhaag-button-medium-size-padding-inline: var(--denhaag-button-padding-inline);
   --denhaag-button-large-size-padding-block: 10px;
   --denhaag-button-large-size-padding-inline: 20px;
+}
```
Robbert, line 31: Ideally this `:root {}` declaration would be moved to the new theme package, and the button component itself does not come with any The Hague specific styles. Or to put it differently: component CSS files are best kept side-effect free, only affecting the page when any of the specified class names are applied to the HTML.

bddjong: This will be discussed with @mjcarsjens

GewoonMaarten (approved):

GewoonMaarten (approved):

src/styles/BaseStyles/src/button.module.css
```diff
@@ -0,0 +1,29 @@
...
+  --denhaag-button-secondary-action-hover-color: var(--denhaag-green-4);
+  --denhaag-button-secondary-action-hover-border-color: var(--denhaag-green-4);
+  --denhaag-button-secondary-action-disabled-background-color: var(--denhaag-white);
+  --denhaag-button-secondary-action-disabled-color: var(--denhaag-grey-2);
+  --denhaag-button-secondary-action-disabled-border-color: var(--denhaag-grey-2);
+
+  --denhaag-button-medium-size-padding-block: 8px;
+  --denhaag-button-medium-size-padding-inline: 16px;
+  --denhaag-button-large-size-padding-block: 10px;
+  --denhaag-button-large-size-padding-inline: 20px;
```
bddjong, line 28: Some of these tokens have a value that refer to a design system token, and some that refer to a very The Hague brand specific token (such as `--denhaag-ocher-4`). What would your advice be for this? @Robbert 

I've split the brand tokens into a specific package, which means that for other developers, these tokens have no value. Is that sufficient? I can think of a few ways to approach this:
- Do not fill or define the button design tokens at all. This means that the external developer _must_ define all the tokens before they can use a component.
- Do not fill the button design tokens if it's not essential styling. This means that the padding tokens will continue to have a value, but most color tokens will not be defined.
- Define all the tokens, giving them a generic value (such as grey tones). Although it's clean to read and might allow for a non-branded version of the storybook, it has pretty much no value to the users, because they won't be using a grey component library.
- Keep this implementation, where some tokens values are brand specific, but the tokens they refer to don't exist if the user isn't importing the brand package.

src/components/Button/src/button.module.css
```diff
@@ -0,0 +1,94 @@
...
+  background-color: var(--denhaag-button-secondary-action-disabled-background-color);
+  color: var(--denhaag-button-secondary-action-disabled-color);
+  border-color: var(--denhaag-button-secondary-action-disabled-border-color);
+}
+
+/* Size specific styling */
+.denhaag-button--medium {
+  padding-inline: var(--denhaag-button-medium-size-padding-inline);
+  padding-block: var(--denhaag-button-medium-size-padding-block);
+  font-size: var(--denhaag-fontsize-base);
```
Robbert, line 65: I would expect the variable to be:
`--denhaag-font-size-base` instead of:
`--denhaag-fontsize-base`

src/components/Button/src/button.module.css
```diff
@@ -0,0 +1,94 @@
+.denhaag-button {
+  font-family: var(--denhaag-font-family);
+  border-radius: var(--denhaag-border-radius);
```
Robbert, line 3: @bddjong For these settings you could introduce design tokens that are common across all form inputs, instead of the generic `font-family` design token from your brand. For example:

```css
:root {
  --denhaag-form-input-font-family: var(--denhaag-font-family);
  --denhaag-form-input-border-radius: var(--denhaag-border-radius);
}
```

```css
.denhaag-button {
  font-family: var(--denhaag-form-input-font-family);
  border-radius: var(--denhaag-form-input-border-radius);
}
```

You could consider "form-input" an 'abstract' component, like the parent component of the button and text input components. You could choose to not define it and rely on system fonts, or use a custom font. That way the component no longer directly relies on your brand design tokens.

src/styles/BaseStyles/src/tokens-button.css
```diff
@@ -0,0 +1,29 @@
+:root {
+  --denhaag-button-border-radius: 3px;
+  --denhaag-button-border-width: 0.0625rem;
+  --denhaag-button-focus-border-color: var(--denhaag-ocher-4);
```
Robbert, line 4: Using `var(--denhaag-ocher-4)` and the other brand design tokens in this file seems fine to me.

src/styles/DenHaagBrand/src/denhaag-brand.module.css
```diff
@@ -1,5 +1,10 @@
 :root {
-  /* Colors */
+  font-size: 16px;
```
Robbert, line 2: @bddjong I recall a discussion we had about avoiding `px` values... was an hardcoded `16px` the conclusion? Would love to hear more about this.

src/components/Button/src/button.module.css
```diff
@@ -0,0 +1,94 @@
...
+.denhaag-button--medium {
+  padding-inline: var(--denhaag-button-medium-size-padding-inline);
+  padding-block: var(--denhaag-button-medium-size-padding-block);
+  font-size: var(--denhaag-fontsize-base);
+
+}
+
+.denhaag-button--medium.denhaag-button--secondary-action {
+  padding-inline: calc(var(--denhaag-button-medium-size-padding-inline) - var(--denhaag-button-border-width));
+  padding-block: calc(var(--denhaag-button-medium-size-padding-block) - var(--denhaag-button-border-width));
```
Robbert, line 71: I would be curious to check: is this `calc()` still necessary when you would use `box-sizing: border-box`?

src/components/Button/src/button.module.css
```diff
@@ -0,0 +1,94 @@
...
+}
+
+.denhaag-button--large {
+  padding-inline: var(--denhaag-button-large-size-padding-inline);
+  padding-block: var(--denhaag-button-large-size-padding-block);
```
Robbert, line 76: With this code, teams that don't need a different padding for all the various button sizes, still are required to specify the `--denhaag-button-large-size-padding-inline` token. Providing the regular padding as fallback value would make it much more convenient to get started and reduce maintenance. Another benefit of this approach is that introducing new variables wouldn't be a breaking change.

For example:

```css
.denhaag-button--large {
   padding-inline: var(--denhaag-button-large-size-padding-inline, var(--denhaag-button-padding-inline));
   padding-block: var(--denhaag-button-large-size-padding-block, var(--denhaag-button-padding-block));
}
```

src/styles/BaseStyles/src/button.module.css
Robbert: Indeed, for colors I don't feel it is feasible to provide a usable fallback value right here in the component. It is possible to pick a limited set of common design tokens for meaningful colors, that you consistently use as fallback value. That way it is easy to quickly apply colors from your brand color palette to these meanings, and quickly achieve a consistent styling.

This page contains a couple of examples for inspiration:
 https://designsystem.wesaynih.com/design-token/color/

Examples:
- success / error
- valid / invalid
- danger / warning / safe
- positive amount / negative amount
- primary action / secondary action / disabled action / destructive action / publication action
- separator color
- text color
- non-essential text color

src/components/Button/src/button.module.css
bddjong: That's a nice touch! Thanks for the suggestion, I'll be sure to add that.

src/components/Button/src/button.module.css
bddjong: I believe I actually tested this and found that the sizing was not coming out correct, but it's worth a double check.

src/styles/DenHaagBrand/src/denhaag-brand.module.css
bddjong: I actually kept the font-size as a hardcoded px because I was under the impression that it would remove ambiguity about the font size that all other rem values are based on. In addition, I don't believe the base font size is necessarily identical in all browsers, so setting it in the Den Haag brand package would enforce the desired scale. 

If this is bad practice I see no problem with removing it!

src/components/Button/src/button.module.css
bddjong: This is where the border between brand design tokens and design system tokens gets vague for me.

The idea of the `--denhaag-font-family` design token is that it's the default token for all font's in the design system, not just a token that only The Hague uses. Although I agree it will be useful to have layered tokens that get gradually get more specific. `--denhaag-font-family` > `--denhaag-form-input-font-family` > `--denhaag-button-font-family`.

Robbert: Yeah, so for Utrecht we now have three directories for design tokens:

- `brand/`
- `common/`
- `component/`

The `--denhaag-font-family` sibling there is `--utrecht-document-font-family` inside `common/`:

https://github.com/nl-design-system/utrecht/blob/336a6f1b9e6c9dc9305381327ae99404ae9601ef/proprietary/design-tokens/common/document.css#L7-L9

You can see it as default value for `paragraph` here, and we will be using it as fallback value for probably all rich text block level elements, but perhaps not for "native" elements such as `<select>`.

https://github.com/nl-design-system/utrecht/blob/240605699bea719d32a7cc68e35f49c99da04c17/src/paragraph/index.css#L6-L7

Robbert: So about the brand design tokens:

- `brand/` contains brand fonts, the brand color palette etc. It can contain fancy brand specific names such as "klm-skyblue" or `military-green` or something like that. Values in this level are not likely to change for a dark mode for example, or for a specific theme or a responsive version.
- `common/` and `component/` is still an expression of your brand, it just applies design decisions and brand values to the design system and the components. You might have different themes for this even for one brand, such as light-on-dark and dark-on-light, dark theme, responsive typography etc.

The values from `brand/` are likely managed by a marketing team with will likely have a different governance process than the application of those values in `common/` and `component/`.

These design tokens together are the theme for The Hague.

Why is brand a separate thing then? First of all, to encourage limiting the choices to stay within your brand parameters and to make it obvious you are going outside of the defined boundaries of your brand when you use a custom hex-color code for example. Second of all, because these brand design tokens can be useful for many projects in your organisation, and it is important to have a single source of truth. Every The Hague project should be able to use those tokens, regardless of wether they use NL Design System or your Material-UI components.

Eventually the design tokens can be made available in other formats than CSS inside a root selector, for that purpose of applying the tokens widely and serving as a single source of truth. For that we are exploring the use of [Style Dictionary](https://amzn.github.io/style-dictionary/) JSON files, you can get an impression of that in the new [amsterdam-design-tokens](https://github.com/nl-design-system/amsterdam-design-tokens/tree/main/proprietary/design-tokens/src) repository where we just started experimenting.

src/styles/DenHaagBrand/src/denhaag-brand.module.css
Robbert: I think allowing ambiguity of the base font size is actually a goal in itself. The folks who change the default font size of their browsers, are likely to need it to be able to read your texts! Why else would they have changed it?

More info: [Pixels vs. Relative Units in CSS: why it’s still a big deal](https://www.24a11y.com/2019/pixels-vs-relative-units-in-css-why-its-still-a-big-deal/)

src/components/Button/src/Button.stories.tsx
```diff
@@ -1,20 +1,20 @@
-import React from 'react';
-// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
-import { Meta, Story } from '@storybook/react/types-6-0';
-import Button, { ButtonProps } from '.';
-import pkg from '../package.json';
+import React from "react";
```
DanielvanVliet, line 1: Check your prettier plugin configuration, this should be single quotes

bddjong: Thanks! I fixed my config but didn't rerun it for that file

.stylelintrc.json
```diff
@@ -30,7 +30,7 @@
     "alpha-value-notation": ["percentage"],
     "hue-degree-notation": ["number"],
     "color-function-notation": ["modern"],
-    "length-zero-no-unit": [true],
+    "length-zero-no-unit": [true, "custom-properties"],
```
Robbert: I'm not sure if this is a valid configuration in your version of `stylelint` already, but just today I ran into the issue that they've changed the configuration API for `custom-properties` in their latest release:

https://github.com/nl-design-system/rijkshuisstijl-design-tokens/blob/d6add187b5bf3f8398374e85e07322d75e7c7ae5/.stylelintrc.json#L33

If it is compatible with the old version, you could already use the new syntax instead.

bddjong: That explains a lot! We've been messing around with that stylelint rule since it was causing an error on some branches. I'm guessing this is the cause.

Robbert: Yeah makes sense! So either you can encourage all developers to install locally with `yarn --frozen-lockfile` which is the yarn variant of `npm ci`.

The other thing you could do is to avoid depending on unreliable version ranges, because it requires the developers of every dependency to follow semantic versioning perfectly, which is unrealistic.

In `npm` projects I usually configure `save-exact=false` in the `.npmrc` file, and I think the `.yarnrc` equivalent is `save-prefix ""`.

In any case I would suggest to use `yarn --frozen-lockfile` instead of just `yarn` in your GitHub workflows:
https://github.com/Gemeente-DenHaag/denhaag-component-library/blob/d6dcf57560e3c98f18a3cba64ceb2936803ef5fd/.github/workflows/cicd.yml#L28

This should help achieving reproducible builds.

bddjong: I'd like @mjcarsjens input on this. I recall the current npm config being a pretty well discussed choice. I don't (yet :D) have the experience to tell whether or not setting save-prefix is beneficial for us without looking into it more deeply. I did find a github issue that weighs the benefits for making it a defualt (https://github.com/yarnpkg/yarn/issues/4147) 

In any case, the linter error has been fixed, thanks!

src/styles/BaseStyles/src/button.module.css
bddjong: Will be implementing meaningful color tokens. I think they would belong in the common tokens package

src/components/Button/src/button.module.css
```diff
@@ -0,0 +1,94 @@
...
+  background-color: var(--denhaag-button-primary-action-background-color);
+  color: var(--denhaag-button-primary-action-color);
+}
+
+.denhaag-button--primary-action.denhaag-button--hover, .denhaag-button--primary-action:hover {
+  background-color: var(--denhaag-button-primary-action-hover-background-color);
+  color: var(--denhaag-button-primary-action-hover-color);
+}
+
+.denhaag-button--primary-action.denhaag-button--disabled {
```
mjcarsjens, line 36: @Robbert I'm personally quite used to styling disabled with `.denhaag-button--primary-action[disabled]`. Looking at it from the perspective of reusable CSS which can be used in a default HTML project, how do you feel about adding this rule here? Is it from your perspective a bad practice to create selectors with attributes?

src/components/Button/src/index.tsx
```diff
@@ -1,18 +1,17 @@
...
    * Simple click handler
    */
   onClick?: () => void;
-
-  /**
-   * Color for the component
-   */
-  color?: 'primary' | 'secondary' | 'default';
-
   /**
```
mjcarsjens, line 15: Do we have a use-case for the href attribute here? From my personal point of view to create a good UX you need a clear visual distinction between buttons which perform actions on a page and links which redirect users to another page/piece of content. I always considered it bad design when a link is visually represented as a button, how do you feel about this? Is it necessary to have this attribute here? (@bddjong @Robbert)

src/components/Button/src/index.tsx
```diff
@@ -32,19 +31,47 @@ export interface ButtonProps extends BaseProps {
...
+      muiVariant = 'contained';
+      break;
+    case 'secondary-action':
+      muiVariant = 'outlined';
+      break;
+  }
+
+  return (
+    <StylesProvider injectFirst>
+      <MaterialButton {...props} classes={classes} className={sizeClass} variant={muiVariant} disableRipple>
```
mjcarsjens, line 70: We still pass all props directly into the Material button. Now this is fine for projects using typescript, but what happens when a user uses our components in a default JS react project? Will his application compile when attributes other than the ones specified in our typing are provided? Because if so we are still simply exposing all possible MaterialUI button props and in that case we should only supply the props which are desirable to us to the MaterialButton.

src/components/Button/src/button.module.css
mjcarsjens: And a quick follow-up question, the `disabled` attribute does this automatically so I don't think this is necessary but do we need `pointer-events: none` here?

Robbert: I agree that in most other projects I would prefer the "standards" way, so in that case I would go with the `:disabled` CSS selector which in turn looks at the `[disabled]` attribute of certain elements.

For maximum reusability, and to allow us to create Storybook stories for every state (even including `--hover`), I am currently leaning towards offering BEM modifiers for every state. The CSS selector that makes assumptions about what attributes exist could be in a separate file, as I mentioned the other day. You can have a look in this PR for the disabled state in Utrecht, where we have the `button:disabled` selector in `html.scss` whereas the BEM class names are defined in another `.css` file.

https://github.com/nl-design-system/utrecht/pull/60

`pointer-events: none` is not necessary, indeed.

A relevant interesting article about this exact subject was published a couple of days ago:
https://css-tricks.com/making-disabled-buttons-more-inclusive/

src/components/Button/src/index.tsx
Robbert: I agree that is not ideal UX, you could start with excluding the `href` attribute for starters, and consider it when the use case arises.

The nice thing about having the BEM class names, is that is easy to copy over the existing styling to a hypothetical Link Button component.

src/components/Button/src/index.tsx
Robbert: Agreed! Avoid committing to a larger API surface than strictly necessary. Removing things is so much harder than adding things, and results in grumpy developers that were depending on those APIs.

src/components/Button/src/button.module.css
bddjong: We've implemented the separation between proprietary, common and component tokens. The hierarchical token structure will be implemented separately when we've added the design for more input components. (Trying to lock this PR down ;P)

src/styles/DenHaagBrand/src/denhaag-brand.module.css
bddjong: Fixed

src/components/Button/src/index.tsx
ogjtech: @Robbert @mjcarsjens the button component has an `onClick` property which contains a method that, surprise surprise, gets called when the button is clicked. Wouldn't it be better to keep the Button's API consistent and do any and all expected functionality from this button click in the onClick method?

Specifying separate properties for each and every use case of a component can and will lead to an insanely bloated API in my opinion.

src/components/Button/src/index.tsx
ogjtech: I don't see how this exposes all possible props of the underlying MaterialUI component? `{...props}` applies every given prop specified in the Properties to the component, no more no less. I am unsure of what happens, if anything, when these components are rendered in JS instead of TS, but I cannot imagine it would fundamentally change the way parameters and/or properties are supplied to a component?

bddjong: > I don't see how this exposes all possible props of the underlying MaterialUI component? `{...props}` applies every given prop specified in the Properties to the component, no more no less. I am unsure of what happens, if anything, when these components are rendered in JS instead of TS, but I cannot imagine it would fundamentally change the way parameters and/or properties are supplied to a component?

I think Maarten was referring to the fact that js wouldn't have type safety, and therefore a user could provide a material UI prop which we don't define in our interface, which would then be passed on to the underlying Material component. I'll be removing the {...props}

mjcarsjens:
This is indeed what I meant, sorry if I wasn't clear in my explanation. To
keep control over how users use our components it is best to not simply
pass on all possible parameters with `{...props}`.

mjcarsjens:
@ogjtech I get what you're saying but you definitely do not want to use the
onClick method when you would want to use a link which looks like a button.
You're creating an absolute a11y nightmare if you are using a html <button>
with on click JS which triggers a routing. When you want something to
behave like a link, just use an <a> tag with an href attribute, and not
something else (which is what the material button does under the hood when
href is supplied).

mjcarsjens:
@bddjong sorry for the shitty formatting, not logged in to GitHub on my
phone so I'm replying to the emails I receive but seems like the formatting
gets hella screwed up this way.

ogjtech:
> When you want something to behave like a link, just use an <a> tag with an href attribute, and not something else (which is what the material button does under the hood when href is supplied).

@mjcarsjens That is indeed the other logical option, just anything other than specifying the `href` prop for the Button is a better option. It is however important to differentiate between expected behaviour of a click on a link or button which is meant to redirect. If redirecting or visiting another link is a process within another process, depending on the exact functionality it could be desirable to redirect through the function (for example redirecting to a 2FA page on login)

ogjtech:
@mjcarsjens @bddjong supplying a shitty href to a Button or linking through onClick results in hella bad a11y nightmares 🤣

mjcarsjens:
@ogjtech Yes you're right about that, there definitely are use-cases for
redirecting in an onClick if more functionality is in the onClick besides
simply redirecting. But seems like we agree about the fact that we would
like to see the href attribute removed, right?

ogjtech:
>  But seems like we agree about the fact that we would like to see the href attribute removed, right?

Yes, it should be and @bddjong already removed in his changes, which have not been pushed yet.

mjcarsjens:
@ogjtech @bddjong noice!

src/components/Button/src/index.tsx
bddjong: I've removed the href

src/components/Button/src/button.module.css
bddjong: It doesn't give me the desired results. I'm not entirely sure. I'd like to investigate why, but preferably not in the scope of this PR.

ogjtech (approved):

GewoonMaarten (approved):

mjcarsjens:
@bddjong seems fine to me to investigate this outside of the scope of this
PR.

mjcarsjens (approved):
Alrighty, finally got home so got my GH password from my keepass and now I can properly comment here. As all comments I left have been handled and as I have no further comments looking at this PR I approve :)

mjcarsjens:
🥳🥳🥳

PR #134, opened 2021-04-14
Feature/#79 cards case

joostvanviegen:
This pull request aims to modify the card component to adhere to the designs of the Card Case variant.


src/components/Card/src/Card/card.module.css
```diff
@@ -66,13 +47,15 @@
...
+    padding-block-start: var(--denhaag-card-subtitle-padding);
+    padding-block-end: var(--denhaag-card-subtitle-padding);
+    padding-inline-start: var(--denhaag-card-subtitle-padding);
+    padding-inline-end: var(--denhaag-card-subtitle-padding);
 }
 
 .denhaag-card--case {
   position: relative;
-  background: var(--denhaag-green-2);
+  background: var(--denhaag-card--case-folder-background-color);
```
Robbert, line 58: CSS variables don't typically follow the `--` BEM double-dash syntax.

joostvanviegen:
@Robbert Firefox on my laptop seems to have some issues with the .denhaag-card--case .denhaag-card__background::before.
We heave deduced we can reduce the gap by increasing the height to 25px. For some reason the perspective is also not working on a reload. In order for the perspective to display correctly we have to use the zoom controls in Storybook (which also narrows the gap on the top part). In Chrome everything works as it should.
afbeelding

mjcarsjens:
> 
> 
> @Robbert @mjcarsjens Firefox on my laptop seems to have some issues with the .denhaag-card--case .denhaag-card__background::before.
> We have deduced we can remove the gap by increasing the height to 25px. For some reason the perspective is also not working on a reload. In order for the perspective to display correctly we have to use the zoom controls in Storybook (which also narrows the gap on the top part). In Chrome and Firefox on my desktop everything works as it should.
> What it looks like:
> afbeelding
> How it should look:
> afbeelding

Hmm, this happens to me as well in Firefox on Windows (latest version), and indeed fixed when zooming once. But only on the preview tab, when going to the docs tab directly the hover effect works fine there (http://localhost:6006/?path=/docs/components-cards-card--case). Have you tried using the component outside of storybook to identify whether this happens only in storybook or also in a real environment?

joostvanviegen:
> > @Robbert @mjcarsjens Firefox on my laptop seems to have some issues with the .denhaag-card--case .denhaag-card__background::before.
> > We have deduced we can remove the gap by increasing the height to 25px. For some reason the perspective is also not working on a reload. In order for the perspective to display correctly we have to use the zoom controls in Storybook (which also narrows the gap on the top part). In Chrome and Firefox on my desktop everything works as it should.
> > What it looks like:
> > afbeelding
> > How it should look:
> > afbeelding
> 
> Hmm, this happens to me as well in Firefox on Windows (latest version), and indeed fixed when zooming once. But only on the preview tab, when going to the docs tab directly the hover effect works fine there (http://localhost:6006/?path=/docs/components-cards-card--case). Have you tried using the component outside of storybook to identify whether this happens only in storybook or also in a real environment?
> 
> Thanks for the zooming tip btw, couldn't figure out why I couldn't get perspective working when I looked at this 2 weeks ago :')

I have tried using it in a near-empty create-react-app but couldn't get the arrow icon or any of the styling to work.
Because the card and a lot of it's dependencies are local I have used yarn add ../denhaag-component-library to copy the needed files to the node_modules folder.

mjcarsjens:
> > > @Robbert @mjcarsjens Firefox on my laptop seems to have some issues with the .denhaag-card--case .denhaag-card__background::before.
> > > We have deduced we can remove the gap by increasing the height to 25px. For some reason the perspective is also not working on a reload. In order for the perspective to display correctly we have to use the zoom controls in Storybook (which also narrows the gap on the top part). In Chrome and Firefox on my desktop everything works as it should.
> > > What it looks like:
> > > afbeelding
> > > How it should look:
> > > afbeelding
> > 
> > 
> > Hmm, this happens to me as well in Firefox on Windows (latest version), and indeed fixed when zooming once. But only on the preview tab, when going to the docs tab directly the hover effect works fine there (http://localhost:6006/?path=/docs/components-cards-card--case). Have you tried using the component outside of storybook to identify whether this happens only in storybook or also in a real environment?
> > Thanks for the zooming tip btw, couldn't figure out why I couldn't get perspective working when I looked at this 2 weeks ago :')
> 
> I have tried using it in a near-empty create-react-app but couldn't get the arrow icon or any of the styling to work.
> Because the card and a lot of it's dependencies are local I have used yarn add ../denhaag-component-library to copy the needed files to the node_modules folder.
> We might want to try testing more of the components which have our own styling, @bddjong thinks the button might have the same problems, haven't had time to test tough.

I'll look into this tonight, quite curious now whether they are actually usable atm. Wrt the card case, am I correct in assuming the perspective bug the only problem left? Might wanna try setting perspective in the transform instead of as a css property on its own (e.g. `transform: perspective(1800px) scaleX(..%)`.

bddjong:
@mjcarsjens @joostvanviegen For the button, I'll do an integration test with the PWA, and a non-typescript project by publishing to a local npm registry with verdaccio.

src/components/Card/src/Card/card.module.css
joostvanviegen: Allright, have made some changes to the variable names

src/components/Button/button.module.css
```diff
@@ -0,0 +1,114 @@
...
+  background-color: var(--denhaag-button-secondary-action-disabled-background-color);
+  color: var(--denhaag-button-secondary-action-disabled-color);
+  border-color: var(--denhaag-button-secondary-action-disabled-border-color);
+}
+
+/* Size specific styling */
+.denhaag-button--medium {
+  padding-inline: 16px;
+  padding-block: 8px;
+  font-size: var(--denhaag-fontsize-base);
```
bddjong, line 85: The correct token is `--denhaag-font-size-...`, this needs to be fixed on a couple of other lines too.

src/components/Card/src/Card/Card.tsx
```diff
@@ -0,0 +1,102 @@
...
+  CardContent,
+  CardActions
+} from '../';
+import { Icon } from '@material-ui/core';
+
+export interface CardProps extends BaseProps {
+  /**
+   * Simple click handler
+   */
+  onClick?: () => void;
```
bddjong, line 26: This should contain a React event. Take a look at the changes @DanielvanVliet made to the IconButton props.

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
+:root {
+    --denhaag-card-arrow-left: 0;
+    --denhaag-card-arrow-position: relative;
+    --denhaag-card-arrow-right: 28px;
+}
```
bddjong, line 5: Why are these tokens not in the component token files?

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+
+.denhaag-card__arrow {
+    left: var(--denhaag-card-arrow-left);
+    position: var(--denhaag-card-arrow-position);
+    right: var(--denhaag-card-arrow-right);
+}
+
+.denhaag-card__title {
+    color: var(--denhaag-black);
```
bddjong, line 41: We shouldn't be referencing to proprietary styles in the component CSS. It's probably a good idea to add a token for the title font color.

src/components/Button/button.module.css
```diff
@@ -0,0 +1,114 @@
+:root {
```
bddjong, line 1: This must be a remainder from a merge with button. This file now exists under Button/src/, you should probably remove this one.

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+
+.denhaag-card--case:hover::before, .denhaag-card--case:focus::before {
+  top: 12px;
+  transform:rotate(-2.45deg);
+  transform-origin: right;
+}
+
+.denhaag-card--case:hover .denhaag-card__background, .denhaag-card--case:focus .denhaag-card__background {
+  transform: rotateX(-20deg);
+}
```
bddjong, line 180: To quote @mjcarsjens, No newline :'(

src/components/Card/src/Card/mui-override.css
```diff
@@ -0,0 +1,13 @@
...
+  overflow: unset;
+}
+
+.denhaag-card--hover, .denhaag-card:hover {
+  box-shadow: none;
+}
+
+.denhaag-card--focus, .denhaag-card:focus {
+  box-shadow: none;
+}
```
bddjong, line 13: newline

src/styles/Components/src/tokens-card.css
```diff
@@ -0,0 +1,34 @@
...
+    --denhaag-card-subtitle-font-size: var(--denhaag-font-size-base);
+    --denhaag-card-subtitle-line-height: 24px;
+    --denhaag-card-subtitle-font-weight: 500;
+    --denhaag-card-subtitle-padding: 0;
+    --denhaag-card-title-font-size: var(--denhaag-font-size-xl);
+    --denhaag-card-title-line-height: 28px;
+    --denhaag-card-title-font-weight: bold;
+    --denhaag-card-title-padding: 0;
+
+    --denhaag-card-case-border: 0;
```
Robbert, line 23: `denhaag-card-case-border` → `denhaag-card-case-border-width`

src/styles/Components/src/tokens-card.css
```diff
@@ -0,0 +1,34 @@
...
+    --denhaag-card-height: 181px;
+    --denhaag-card-width: 352px;
+    --denhaag-card-padding-block: 24px;
+    --denhaag-card-padding-inline: 24px;
+
+    --denhaag-card-subtitle-color: var(--denhaag-grey-4);
+    --denhaag-card-subtitle-font-size: var(--denhaag-font-size-base);
+    --denhaag-card-subtitle-line-height: 24px;
+    --denhaag-card-subtitle-font-weight: 500;
+    --denhaag-card-subtitle-padding: 0;
```
Robbert, line 17: Could we specify which axis for padding this should apply to? (Inline/Block)

src/components/Card/src/tokens-card.css
```diff
@@ -0,0 +1,34 @@
...
+    --denhaag-card-case-padding-block-start: 20px;
+    --denhaag-card-case-padding-block-end: 24px;
+    --denhaag-card-case-width: 352px;
+
+    --denhaag-card-case-text-color: var(--denhaag-green-3);
+    --denhaag-card-case-text-active-color: var(--denhaag-green-4);
+    --denhaag-card-case-primary-background-color: var(--denhaag-green-1);
+    --denhaag-card-case-secundary-background-color: var(--denhaag-green-2);
+    --denhaag-card-case-paper-color: var(--denhaag-white);
+}
```
Robbert, line 34: I suspect this file is obsolete now, I suppose it has been replaced by the one in `src/styles/`.

src/components/Card/src/CardHeader/mui-override.css
```diff
@@ -0,0 +1,12 @@
...
+  transition: none;
+}
+
+.denhaag-card--hover, .denhaag-card:hover {
+  box-shadow: none;
+}
+
+.denhaag-card--focus, .denhaag-card:focus {
+  box-shadow: none;
+}
```
Robbert, line 12: I really like that you have separated the override styles from the rest!

src/components/Card/src/CardHeader/card-header.css
```diff
@@ -0,0 +1,7 @@
+.denhaag-card__header {
+    --denhaag-card__header-font-style: normal;
+    --denhaag-card__header-font-weight: bold;
+    --denhaag-card__header-font-size: 16px;
+    
+    --denhaag-card__header-line-height: 24px;
```
Robbert, line 6: `__` in a custom property should probably be `-`.

src/styles/Components/src/tokens-card.css
```diff
@@ -0,0 +1,34 @@
...
+    --denhaag-card-case-border: 0;
+    --denhaag-card-case-height: 200px;
+    --denhaag-card-case-padding-block-start: 20px;
+    --denhaag-card-case-padding-block-end: 24px;
+    --denhaag-card-case-width: 352px;
+
+    --denhaag-card-case-text-color: var(--denhaag-green-3);
+    --denhaag-card-case-text-active-color: var(--denhaag-green-4);
+    --denhaag-card-case-primary-background-color: var(--denhaag-green-1);
+    --denhaag-card-case-secundary-background-color: var(--denhaag-green-2);
```
Robbert, line 32: `secundary` → `secondary`

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+    line-height: var(--denhaag-card-subtitle-line-height);
+    padding-block-start: var(--denhaag-card-subtitle-padding);
+    padding-block-end: var(--denhaag-card-subtitle-padding);
+    padding-inline-start: var(--denhaag-card-subtitle-padding);
+    padding-inline-end: var(--denhaag-card-subtitle-padding);
+}
+
+.denhaag-card--case {
+  position: relative;
+  background: var(--denhaag-card-case-secundary-background-color);
```
Robbert, line 62: `secundary` → `secondary`

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
+:root {
+    --denhaag-card-arrow-left: 0;
+    --denhaag-card-arrow-position: relative;
+    --denhaag-card-arrow-right: 28px;
```
Robbert, line 4: Is this `:root` selector still supposed to be here?

src/components/Card/src/CardActions/CardAction.tsx
```diff
@@ -0,0 +1,30 @@
...
+  disableSpacing?: boolean;
+}
+
+/**
+ * Primary UI component for user interaction
+ */
+export const CardActions: React.FC<CardActionProps> = (props: CardActionProps) => {
+  return (
+    <StylesProvider injectFirst>
+      <MaterialCardActions classes={cardActionClasses} {...props}>{props.children}</MaterialCardActions>
```
bddjong, line 22: `{...props}`
We don't want to blindly pass props to the underlying material UI component. This potentially exposes MUI API which we don't wish to support. See this discussion for more details: https://github.com/Gemeente-DenHaag/denhaag-component-library/pull/133#discussion_r637213463

src/components/Card/src/CardActions/bem-mapping.jsx
```diff
@@ -0,0 +1,3 @@
+export const cardActionClasses = {
+    root: 'denhaag-card-actions'
```
bddjong, line 2: I'm not 100% sure, but I think according to BEM, actions is an element of the denhaag-card. So it would be `denhaag-card__actions`

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+}
+
+.denhaag-card--case .denhaag-card__subtitle {
+  position: relative;
+  color: var(--denhaag-card-case-text-color);
+  font-weight: 500;
+  padding-block: var(--denhaag-card-subtitle-padding);
+  padding-inline: var(--denhaag-card-subtitle-padding);
+  /*flex-grow: 1;/*
+  /* height: 44px; */
```
bddjong, line 165: Do we need this commented out code?

src/components/Card/src/CardActions/card-actions.css
```diff
@@ -0,0 +1,8 @@
+:root {
+    --denhaag-card-actions-padding-top: 8px;
```
bddjong, line 2: This root declaration should probably be moved to the component tokens folder

src/components/Card/src/tokens-card.css
```diff
@@ -0,0 +1,34 @@
+:root {
+    --denhaag-card-background-color: var(--denhaag-white);
+
+    --denhaag-card-border-radius: 3px;
```
bddjong, line 4: You can use the common border radius token for this

bddjong (changes requested):
Few comments on the tokens and styling, looking good overal!

src/components/Card/src/Card/Card.tsx
```diff
@@ -0,0 +1,102 @@
...
+        </div>
+      </MaterialCard>
+    </StylesProvider>
+  )
+};
+
+/**
+ * Default export for Card
+ */
+export default Card;
```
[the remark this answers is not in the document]
mjcarsjens: @bddjong you missed this one

src/components/Card/src/CardActions/bem-mapping.jsx
mjcarsjens: Yea I agree, as this subcomponent is always and only used as part of a Card this would be an element in the card block.

src/components/Card/src/CardContent/CardContent.tsx
```diff
@@ -18,7 +21,11 @@ export interface CardContentProps extends BaseProps {
  * Primary UI component for user interaction
  */
 export const CardContent: React.FC<CardContentProps> = (props: CardContentProps) => {
-  return <MaterialCardContent {...props}>{props.children}</MaterialCardContent>;
+  return (
+      <StylesProvider injectFirst>
+        <MaterialCardContent classes={CardContentClasses} {...props}>{props.children}</MaterialCardContent>
```
mjcarsjens, line 26: As stated above by @bddjong. we should avoud `{...props}`.

src/components/Card/src/CardContent/mui-override.css
```diff
@@ -0,0 +1,13 @@
+.denhaag-card__content {
+    padding-block: 0 0;
```
mjcarsjens, line 2: Why `0 0` instead just `0` here?

src/components/Card/src/CardContent/mui-override.css
```diff
@@ -0,0 +1,13 @@
...
+    padding-inline: 0;
+
+    display: flex;
+    flex-direction: column;
+    flex-grow: 1;
+    justify-content: space-between;
+}
+
+.denhaag-card__content:last-child {
+    padding-block: 0 0;
```
mjcarsjens, line 12: As above, why `0 0`?

src/components/Card/src/CardHeader/CardHeader.tsx
```diff
@@ -54,7 +58,12 @@ export interface CardHeaderProps extends BaseProps {
  * Primary UI component for user interaction
  */
 export const CardHeader: React.FC<CardHeaderProps> = (props: CardHeaderProps) => {
-  return <MaterialCardHeader {...props}>{props.children}</MaterialCardHeader>;
+  return (
+    <StylesProvider injectFirst>
+      <MaterialCardHeader {...props}>{props.children}</MaterialCardHeader>
```
mjcarsjens, line 63: `{...props}`

mjcarsjens (dismissed):
Left some comments of my own which should be resolved/discussed, furthermore I agree with all comments left by @bddjong and @Robbert .

mjcarsjens:
@joostvanviegen I also noticed that the focus border around the arrow only shows up for card/case and not for the card/default. Is this intentional?

mjcarsjens:
@joostvanviegen another one: bit of a trickier one: you added `tabindex="0"` to make the element focusable by keyboard, and a `onClick` prop to make it clickable. I have a few issues with this though:

- I feel like in our use-case a card would almost always be a link which we would want to define as an `<a>` tag with an `href` instead of a `onClick` function. Usually you would not want to wrap the whole card in an `<a>`-tag as this creates a bit of a screenreader nightmare though. I would personally prefer placing the title (and maybe subtitle) in an `<a>` tag with an `href` attribute. Then add a JS `focus` and `blur` listener which adds the `denhaag-card--focus` class to the card providing the focus border. @Robbert how do you look at this?

- And when deciding to work with the `onClick` function: you have made the card focusable by adding tabIndex, so at least it is reachable by tabbing through the page, however nothing in the HTML indicates that this element is actually clickable. Sure, tabindex makes it focusable but the fact that there is a JS 'onClick' listener here can not be discerned from the HTML as it is simply a bunch of `<div>` and `<p>` tags. This means screenreaders and other a11y tools will not actually detect the card as something the user can click. This can be fixed by either adding a `<button>` tag (which is preferred) somewhere (preferably again with little content, so for example only around the title and subtitle). If for some reason this would not be possible you can add the `role="button"` attribute to a not by default clickable html tag (so in this case a `div`) which will identify it to screenreaders as something the user can click. Using the `role` attribute for this should however only be considered in scenarios where it is impossible to use the existing HTML tag for whatever reason. 

@Gemeente-DenHaag/mdh-rysst A quick reminder to keep an eye out for a11y problems with our component implementations!

src/components/Card/src/index.ts
```diff
@@ -1,8 +1,8 @@
...
 export default Card;
-export * from './Card';
-export * from './CardActions';
-export * from './CardContent';
-export * from './CardHeader';
-export * from './Paper';
+export * from './Card/Card';
+export * from './CardHeader/CardHeader';
+export * from './CardActions/CardAction';
+export * from './CardContent/CardContent'
```
mjcarsjens, line 7: Do we actually want to export these? I agree this is the case for `CardProps` (`export *  from './Card/Card'`) and maybe `Paper` as that might be seen as a own component (even though it is quite useless atm). I don't really see a use-case why we want to expose all the components which are only meant to be used as part of the Card though.

src/components/Card/src/Card/card.css
joostvanviegen: We don't

src/components/Card/src/Card/Card.tsx
```diff
@@ -0,0 +1,102 @@
...
+            <div className='denhaag-card--case__text__wrapper'>
+              <Typography classes={titleClasses} component='p'>
+                {props.title}
+              </Typography>
+              <Typography classes={subtitleClasses} component='p'>
+                {props.subTitle}
+              </Typography>
+            </div>
+            <CardActions disableSpacing={true}>
+              <Typography component='div' classes={subtitleClasses}>
```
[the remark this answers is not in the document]
joostvanviegen: I have added a nice time html5 tag around the date.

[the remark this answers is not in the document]
joostvanviegen: @mjcarsjens  A datetime component could be a good idea but i'm not sure if it is worth the overhead.

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+    border-radius: var(--denhaag-card-border-radius);
+    border-width: var(--denhaag-card-border-width);
+    border-color: var(--denhaag-card-border-color);
+    border-style: solid;
+    box-sizing: border-box;
+    height: var(--denhaag-card-height);
+    padding-inline: var(--denhaag-card-padding-inline);
+    padding-block: var(--denhaag-card-padding-block);
+    flex-direction: column;
+    display: flex;
```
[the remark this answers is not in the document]
joostvanviegen: The idea was for the css files to have alphabetic ordering, but I seem to not have done this thoroughly enough. I have made an issue to add such a linter plugin (#220 ).

src/components/Card/src/index.ts
joostvanviegen: Good point, I have made some changes to only export the card and paper components for now.

src/styles/Components/src/tokens-card.css
joostvanviegen: done

src/styles/Components/src/tokens-card.css
joostvanviegen: Right now it applies to all padding of the subtitle i'm not sure if we want to have individual tokens for all of the padding of the subtitle. @mjcarsjens what do you think?

mjcarsjens:
@joostvanviegen FYI, found out that you were correct in your observation that the outline had a slight curve to it in Firefox.

`From Firefox 88, outline now follows the shape created by border-radius automatically`. This is still only implemented in Firefox and a fairly new feature (April 19th of this year) which is why I hadn't come across it yet 😅

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+.denhaag-card__arrow {
+    left: var(--denhaag-card-arrow-left);
+    position: var(--denhaag-card-arrow-position);
+    right: var(--denhaag-card-arrow-right);
+}
+
+.denhaag-card__title {
+    color: var(--denhaag-black);
+    padding-block-start: var(--denhaag-card-title-padding);
+    padding-block-end: var(--denhaag-card-title-padding);
```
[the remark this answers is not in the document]
joostvanviegen: @mjcarsjes From my understanding the compatibility of `padding-block-start` and `padding-block-end` is higher 91.82% vs just using `padding-block` 68.43% source https://caniuse.com/?search=padding-block

src/components/Card/src/Card/card.css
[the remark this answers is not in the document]
joostvanviegen: Have resorted the properties in alphabetic ordering for now.

src/components/Card/src/Card/card.css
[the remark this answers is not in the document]
Robbert: @joostvanviegen Yeah, that's why I always use the terribly long versions instead too.

src/components/Card/src/CardHeader/card-header.css
joostvanviegen: Done

src/components/Card/src/Card/bem-mapping.jsx
```diff
@@ -0,0 +1,27 @@
...
+export const cardPaperClasses = {
+  root: 'denhaag-card__paper'
+}
+
+export const cardCaseClasses = {
+  root: 'denhaag-card denhaag-card--case'
+}
+
+export const cardCaseFocusClasses = {
+  root: 'denhaag-card--case__text__wrapper'
```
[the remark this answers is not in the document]
joostvanviegen: Correct

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+    flex-direction: column;
+    display: flex;
+}
+
+.denhaag-card:focus .denhaag-card--case__text__wrapper {
+  outline: var(--denhaag-focus-border);
+  position: relative;
+}
+
+.denhaag-card:focus {
```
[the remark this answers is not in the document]
joostvanviegen: Done

src/components/Card/src/CardActions/CardAction.tsx
joostvanviegen: Fixed

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+
+.denhaag-card__subtitle {
+    color: var(--denhaag-card-subtitle-color);
+    font-size: var(--denhaag-card-subtitle-font-size);
+    font-weight: var(--denhaag-card-subtitle-font-weight);
+    line-height: var(--denhaag-card-subtitle-line-height);
+    padding-block-start: var(--denhaag-card-subtitle-padding);
+    padding-block-end: var(--denhaag-card-subtitle-padding);
+    padding-inline-start: var(--denhaag-card-subtitle-padding);
+    padding-inline-end: var(--denhaag-card-subtitle-padding);
```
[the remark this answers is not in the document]
joostvanviegen: https://github.com/Gemeente-DenHaag/denhaag-component-library/pull/134#discussion_r648057710

src/components/Card/src/CardHeader/CardHeader.tsx
joostvanviegen: Cardheader has been removed since we don't actually use it in card

src/components/Card/src/Card/card.css
```diff
@@ -0,0 +1,180 @@
...
+}
+
+.denhaag-card--case {
+  position: relative;
+  background: var(--denhaag-card-case-secundary-background-color);
+  padding-block-start: 20px;
+  border-radius: var(--denhaag-card-border-radius);
```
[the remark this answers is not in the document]
joostvanviegen: @mjcarsjens  Not really sure what you meant. But I think it is fixed now.

src/components/Card/src/CardContent/mui-override.css
joostvanviegen: Becausejust 0 won't override the specifity of the padding-bottom material-ui uses for the underlying cardcontent component

src/components/Card/src/CardContent/mui-override.css
joostvanviegen: See above

joostvanviegen:
> @joostvanviegen another one, this time a bit of a trickier one though: you added `tabindex="0"` to make the element focusable by keyboard, and a `onClick` prop to make it clickable. I have a few issues with this though:
> 
>     * I feel like in our use-case a card would almost always be a link which we would want to define as an `<a>` tag with an `href` instead of a `onClick` function. Usually you would not want to wrap the whole card in an `<a>`-tag as this creates a bit of a screenreader nightmare though. I would personally prefer placing the title (and maybe subtitle) in an `<a>` tag with an `href` attribute. Then add a JS `focus` and `blur` listener which adds/removes the `denhaag-card--focus` class to the card providing the focus border, and a JS `click` listener to pass through the click to the `a` tag when the card is clicked. Furthermore you should add a `cursor: pointer;` rule to the `.denhaag-card` CSS to actually transform the mouse into a pointer when the card is hovered to indicate it is clickable as by default this only happens when hovering a `<a>` or `<button>`. @Robbert how do you look at this?
> 
>     * And when deciding to work with the `onClick` function (which again, I don't prefer in this scenario): you have made the card focusable by adding `tabindex`, so at least it is reachable by tabbing through the page, however nothing in the HTML indicates that this element is actually clickable. Sure, tabindex makes it focusable but the fact that there is a `onClick` listener here can not be discerned from the HTML as it is simply a bunch of `<div>` and `<p>` tags. This means screenreaders and other a11y tools will not actually detect the card as something the user can click. This can be fixed by either adding a `<button>` tag (which is preferred) somewhere (preferably again with little content, so for example only around the title and subtitle). If for some reason this would not be possible you can add the `role="button"` attribute to a not by default clickable html tag (so in this case a `div`) which will identify it to screenreaders as something the user can click. Using the `role` attribute for this should however only be considered in scenarios where it is impossible to use the existing HTML tag for whatever reason.
> 
>     * The card is now focusable when no `onClick` is provided because of the `tabindex="0"`: it feels a bit weird to be able to focus something which doesn't actually do anything. Above i mentioned the `cursor: pointer` CSS rule, but this also should only happen when the card is actually clickable. I feel like this component should always have an action (either `button` with an `onClick` or an `a` with an `href`, see above). Right now you can create a card which does nothing because you don't have to provide a `onClick` prop. I feel like we should make it required (either `onClick` or `href`, depending on what we decide based on the discussion which will follow on my comments above).
> 
> 
> @Gemeente-DenHaag/mdh-rysst A quick reminder to keep an eye out for a11y problems with our component implementations!

Haven't gotten around to this one because of the other suggestions / questions but 

> @joostvanviegen I also noticed that the focus border around the arrow only shows up for card/case and not for the card/default. Is this intentional?

Is now fixed

joostvanviegen:
> @joostvanviegen another one, this time a bit of a trickier one though: you added `tabindex="0"` to make the element focusable by keyboard, and a `onClick` prop to make it clickable. I have a few issues with this though:
> 
>     * I feel like in our use-case a card would almost always be a link which we would want to define as an `<a>` tag with an `href` instead of a `onClick` function. Usually you would not want to wrap the whole card in an `<a>`-tag as this creates a bit of a screenreader nightmare though. I would personally prefer placing the title (and maybe subtitle) in an `<a>` tag with an `href` attribute. Then add a JS `focus` and `blur` listener which adds/removes the `denhaag-card--focus` class to the card providing the focus border, and a JS `click` listener to pass through the click to the `a` tag when the card is clicked. Furthermore you should add a `cursor: pointer;` rule to the `.denhaag-card` CSS to actually transform the mouse into a pointer when the card is hovered to indicate it is clickable as by default this only happens when hovering a `<a>` or `<button>`. @Robbert how do you look at this?
> 
>     * And when deciding to work with the `onClick` function (which again, I don't prefer in this scenario): you have made the card focusable by adding `tabindex`, so at least it is reachable by tabbing through the page, however nothing in the HTML indicates that this element is actually clickable. Sure, tabindex makes it focusable but the fact that there is a `onClick` listener here can not be discerned from the HTML as it is simply a bunch of `<div>` and `<p>` tags. This means screenreaders and other a11y tools will not actually detect the card as something the user can click. This can be fixed by either adding a `<button>` tag (which is preferred) somewhere (preferably again with little content, so for example only around the title and subtitle). If for some reason this would not be possible you can add the `role="button"` attribute to a not by default clickable html tag (so in this case a `div`) which will identify it to screenreaders as something the user can click. Using the `role` attribute for this should however only be considered in scenarios where it is impossible to use the existing HTML tag for whatever reason.
> 
>     * The card is now focusable when no `onClick` is provided because of the `tabindex="0"`: it feels a bit weird to be able to focus something which doesn't actually do anything. Above i mentioned the `cursor: pointer` CSS rule, but this also should only happen when the card is actually clickable. I feel like this component should always have an action (either `button` with an `onClick` or an `a` with an `href`, see above). Right now you can create a card which does nothing because you don't have to provide a `onClick` prop. I feel like we should make it required (either `onClick` or `href`, depending on what we decide based on the discussion which will follow on my comments above).
> 
> 
> @Gemeente-DenHaag/mdh-rysst A quick reminder to keep an eye out for a11y problems with our component implementations!

Haven't gotten around to this one because of all the other comments but will do this next week.

src/components/Card/src/Card/Card.tsx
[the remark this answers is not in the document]
mjcarsjens: Looks nice, for now this is fine indeed, if we run into the scenario where we do need a datetime component we can always add it at that point

src/styles/Components/src/tokens-card.css
mjcarsjens: I'm fine with splitting this up into `inline` and `block`, feel like there's a valid usecase for that. You can always just use the same setup as above for `--denhaag-card-wrapper-padding`

mjcarsjens:
Making some nice progress with resolving the comments! lmk if you need help with the a11y comment

joostvanviegen:
> @joostvanviegen another one, this time a bit of a trickier one though: you added `tabindex="0"` to make the element focusable by keyboard, and a `onClick` prop to make it clickable. I have a few issues with this though:
> 
>     * I feel like in our use-case a card would almost always be a link which we would want to define as an `<a>` tag with an `href` instead of a `onClick` function. Usually you would not want to wrap the whole card in an `<a>`-tag as this creates a bit of a screenreader nightmare though. I would personally prefer placing the title (and maybe subtitle) in an `<a>` tag with an `href` attribute. Then add a JS `focus` and `blur` listener which adds/removes the `denhaag-card--focus` class to the card providing the focus border, and a JS `click` listener to pass through the click to the `a` tag when the card is clicked. Furthermore you should add a `cursor: pointer;` rule to the `.denhaag-card` CSS to actually transform the mouse into a pointer when the card is hovered to indicate it is clickable as by default this only happens when hovering a `<a>` or `<button>`. @Robbert how do you look at this?
> 
>     * And when deciding to work with the `onClick` function (which again, I don't prefer in this scenario): you have made the card focusable by adding `tabindex`, so at least it is reachable by tabbing through the page, however nothing in the HTML indicates that this element is actually clickable. Sure, tabindex makes it focusable but the fact that there is a `onClick` listener here can not be discerned from the HTML as it is simply a bunch of `<div>` and `<p>` tags. This means screenreaders and other a11y tools will not actually detect the card as something the user can click. This can be fixed by either adding a `<button>` tag (which is preferred) somewhere (preferably again with little content, so for example only around the title and subtitle). If for some reason this would not be possible you can add the `role="button"` attribute to a not by default clickable html tag (so in this case a `div`) which will identify it to screenreaders as something the user can click. Using the `role` attribute for this should however only be considered in scenarios where it is impossible to use the existing HTML tag for whatever reason.
> 
>     * The card is now focusable when no `onClick` is provided because of the `tabindex="0"`: it feels a bit weird to be able to focus something which doesn't actually do anything. Above i mentioned the `cursor: pointer` CSS rule, but this also should only happen when the card is actually clickable. I feel like this component should always have an action (either `button` with an `onClick` or an `a` with an `href`, see above). Right now you can create a card which does nothing because you don't have to provide a `onClick` prop. I feel like we should make it required (either `onClick` or `href`, depending on what we decide based on the discussion which will follow on my comments above).
> 
> 
> @Gemeente-DenHaag/mdh-rysst A quick reminder to keep an eye out for a11y problems with our component implementations!

I have implemented a fix from the article suggested by @Robbert. @mjcarsjens Would be nice if you could look at it again to see if you agree with this solution.

src/components/Card/src/Card/Card.tsx
```diff
@@ -0,0 +1,110 @@
...
+  return (
+    <StylesProvider injectFirst>
+      <MaterialCard classes={classes} onClick={onClick}>
+        <div className="denhaag-card__wrapper">
+          <div className="denhaag-card__background"></div>
+          <CardContent>
+            <div className="denhaag-card__text-wrapper">
+              <Typography classes={titleClasses} component="p">
+                {title}
+              </Typography>
```
mjcarsjens, line 87: Nice use of the `<a>` tag, looks like this is a solid setup wrt a11y!
Two questions:
- What made you decide to place the HTML of the `<a>`-tag in a variable instead of just right here?
- Might be a bit of a discussion point whether or not a `<p>` HTML tag is actually semantically correct here. As the name implies `<p>` is a `paragraph` tag, used to semantically markup a piece of text as a paragraph. In my eyes a single line of text for a title or a subtitle does not make a paragraph which makes me think this is not actually semantically correct HTML use. Not sure what the best approach would be in this construction though as the `Typography` component is necessary for the styling. My first thought is to use a extra `<div>` for the block, with a `Typography` component which uses a HTML `span`.

src/styles/Components/src/tokens-card.css
```diff
@@ -0,0 +1,56 @@
+:root {
+  --denhaag-card__actions-padding-block-end: 0;
+  --denhaag-card__actions-padding-block-start: 8px;
+  --denhaag-card__actions-padding-inline-end: 0;
+  --denhaag-card__actions-padding-inline-start: 0;
```
mjcarsjens, line 5: You're using `__` in the design tokens here, this should be replaced by a `-`, so for example: `--denhaag-card-actions-padding-block-end`

mjcarsjens (commented):
2 small comments left but once these are resolved I think we can call this one finished, nice work!

src/components/Card/src/Card/Card.tsx
DanielvanVliet: For the first point, I think this helps to make clear where the react Ref is used and we should stick to this.
For the second point, in the future we should refactor CardCase to make use of the  updated Typography component from our library. We can move the discussion to that refactoring.
https://github.com/Gemeente-DenHaag/denhaag-component-library/issues/253

src/styles/Components/src/tokens-card.css
DanielvanVliet: Resolved

bddjong (approved):

GewoonMaarten (approved):

PR #145, opened 2021-04-24
Defined prettifier scripts which run in specified directories only

ogjtech:
This PR defines two shell scripts in the root directory which, using the already defined config and ignore file, run only in either the stories folder or the directories folder. Doing so, running prettifier in a branch does not introduce hundreds of changes in files which would otherwise not be touched by the branch,

### DIFF
--- a/prettify_components.sh
+++ b/prettify_components.sh
@@ -3,7 +3,7 @@
 # Make sure the working directory for this command is equal to "<PROJECT_DIR>/src/components"
 
 # The path to the prettier config file.
-CONFIG_FILE="../../.pretierrc"
+CONFIG_FILE="../../.prettierrc"
 
 # The path to the prettier ignore file.
 IGNORE_FILE="../../.prettierignore"
@@ -13,6 +13,6 @@ IGNORE_FILE="../../.prettierignore"
 # *.tsx = all files which are appended with ".tsx" in the specified directory.
 COMPONENTS_DIR_REGEX="./**/*.tsx"
 
-prettier --config $CONFIG_FILE --ignore-path $IGNORE_FILE all --write $COMPONENTS_DIR_REGEX
+prettier --config $CONFIG_FILE --ignore-path $IGNORE_FILE all --write "$COMPONENTS_DIR_REGEX"
 
 # After this command is run, ensure to open all the components at least once for the changes to actually be applied and registered by the VC diff manager.
--- a/prettify_components.sh
+++ b/prettify_components.sh
@@ -1,0 +1,18 @@
+# Add this as a shell script to your WebStorm configurations
+# and as executable command to VS Code
+# Make sure the working directory for this command is equal to "<PROJECT_DIR>/src/components"
+
+# The path to the prettier config file.
+CONFIG_FILE="../../.pretierrc"
+
+# The path to the prettier ignore file.
+IGNORE_FILE="../../.prettierignore"
+
+# The regex for the exact files to be handled by prettifier.
+# ** = all (sub-)directories in the current working directory.
+# *.tsx = all files which are appended with ".tsx" in the specified directory.
+COMPONENTS_DIR_REGEX="./**/*.tsx"
+
+prettier --config $CONFIG_FILE --ignore-path $IGNORE_FILE all --write $COMPONENTS_DIR_REGEX
+
+# After this command is run, ensure to open all the components at least once for the changes to actually be applied and registered by the VC diff manager.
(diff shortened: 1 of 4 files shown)


prettify_components.sh
```diff
@@ -0,0 +1,18 @@
+# Add this as a shell script to your WebStorm configurations
+# and as executable command to VS Code
+# Make sure the working directory for this command is equal to "<PROJECT_DIR>/src/components"
+
+# The path to the prettier config file.
+CONFIG_FILE="../../.pretierrc"
```
Robbert, line 6: `pretierrc` → `prettierrc`?

Also, personally I prefer to name that file `.prettierrc.json` with the `.json` extension so it has the correct syntax highlighting and is formatted itself by Prettier too, and any accidental syntax errors can be picked up by linting.

prettify_stories.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# The path to the prettier ignore file.
+IGNORE_FILE="../../.prettierignore"
+
+# The regex for the exact files to be handled by prettifier.
+# ** = all (sub-)directories in the current working directory.
+# *.stories.tsx = all files which are appended with ".stories.tsx" in the specified directory.
+STORIES_DIR_REGEX="./**/*.stories.tsx"
+
+prettier --config $CONFIG_FILE --ignor-path $IGNORE_FILE all --write $STORIES_DIR_REGEX
```
Robbert, line 16: `--ignor-path` :rarr `--ignore-path`

Hehe, I think maybe you should test this script first?

prettify_components.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# The path to the prettier ignore file.
+IGNORE_FILE="../../.prettierignore"
+
+# The regex for the exact files to be handled by prettifier.
+# ** = all (sub-)directories in the current working directory.
+# *.tsx = all files which are appended with ".tsx" in the specified directory.
+COMPONENTS_DIR_REGEX="./**/*.tsx"
+
+prettier --config $CONFIG_FILE --ignore-path $IGNORE_FILE all --write $COMPONENTS_DIR_REGEX
```
Robbert, line 16: I'm a bit worried that not escaping it as `"$COMPONENTS_DIR_REGEX"` might lead to "glob expansion".

See [Shellcheck](https://www.shellcheck.net):

Screenshot 2021-04-24 at 14 43 02

prettify_components.sh
ogjtech: In my IDE it already shows as a JSON file. Likely a `.prettier`-file already translates to JSON in the background as it read as such. Also, syntax highlighting seems to work fine for me.
afbeelding

scripts/prettify_components.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Add this as a shell script to your WebStorm configurations
+# and as executable command to VS Code
+# Make sure the working directory for this command is equal to "<PROJECT_DIR>/src/components"
+
+# The path to the prettier config file.
+CONFIG_FILE="../../.prettierrc"
```
GewoonMaarten, line 8: Shouldn't this be just `.prettierrc`?
```suggestion
CONFIG_FILE=".prettierrc"
```

scripts/prettify_components.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# The path to the prettier ignore file.
+IGNORE_FILE="../../.prettierignore"
```
GewoonMaarten, line 11: Same for this one

scripts/prettify_components.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# The regex for the exact files to be handled by prettifier.
+# ** = all (sub-)directories in the current working directory.
+# *.tsx = all files which are appended with ".tsx" in the specified directory.
+COMPONENTS_DIR_REGEX="./**/*.tsx"
+
+prettier --config $CONFIG_FILE --ignore-path $IGNORE_FILE all --write "$COMPONENTS_DIR_REGEX"
```
GewoonMaarten, line 18: If you add yarn then people wont have to install globally.
```suggestion
yarn prettier --config $CONFIG_FILE --ignore-path $IGNORE_FILE all --write "$COMPONENTS_DIR_REGEX"
```

Also I get the error: `[error] No files matching the pattern were found: "all".`

scripts/prettify_components.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
Robbert, line 1: Often I use `#!/bin/bash -e` as shebang, because the default behaviour to continue the script after an error is usually quite undesired.

GewoonMaarten: added 👍

Issue #156, opened 2021-05-01
Separate design tokens and DH brand


bddjong:
Currently, the brand design tokens are part of the base styles package. Brand and component design tokens should be separated.

**Acceptance criteria**
- [x] `brand.module.css` is available in a separate `denhaag-brand-styles` package
- [x] `design-tokens.module.css` is available in the base styles package
- [x] Style packages are moved to separate folder under /src

PR #157, opened 2021-05-01
Fix/#156 separate design tokens from brand

bddjong:
Implements #156 

The Hague brand specific tokens should be separated from component tokens. This PR creates the following folders:
```
src/
  |-styles/
    |-BaseStyles/
    |-DenHaagBrand/
```
It also adds a dev dependency on copyfiles, which is used to bundle the CSS with the compiled typescript in the dist. (I'm pretty sure we want to use webpack or rollup to handle that and optimize the files, but that's out of scope for this issue.) 

Finally, .storybook/preview.js has been changed to import the Den Haag brand css


src/styles/BaseStyles/src/globals.module.css
```diff
@@ -0,0 +1,23 @@
...
+  /*
+   * Base tokens
+   */
+  --denhaag-fontsize-xs: 0.625rem;
+  --denhaag-fontsize-s: 0.75rem;
+  --denhaag-fontsize-base: 1rem;
+  --denhaag-fontsize-lg: 1.25rem;
+  --denhaag-fontsize-xl: 1.5rem;
+  --denhaag-fontsize-2xl: 2rem;
+  --denhaag-fontsize-3xl: 3rem;
```
bddjong, line 11: It's a bit unclear to me whether these should be brand tokens or design system tokens. In my opinion, the font sizes are not very brand specific, and having a fontsize scale defined in the basestyles will make it easier for developers to easily change the font sizes in the component library, without needing to change all component specific fontsize tokens.

GewoonMaarten (approved):

mjcarsjens (approved):
Seems like a solid first setup to me. 

Slightly unrelated comment: @Robbert I'm personally used to also defining margins/border-width/radius etc. in rem instead of using px values. What are your thoughts about the use of rem/em/px values?

Robbert:
The article [Pixels vs. Relative Units in CSS: why it’s still a big deal](https://www.24a11y.com/2019/pixels-vs-relative-units-in-css-why-its-still-a-big-deal/) from 2019 really dives into this subject.

@mjcarsjens My conclusion is that `rem` should be preferred in CSS over `px`, but you should document both because `px` is what you see in Figma and other tools.

PR #158, opened 2021-05-01
Feature/#46 icons

GewoonMaarten:
closes #158


DanielvanVliet (changes requested):
In storybook, the single icon story does not show any icon.
The show code does not seem correct either for single icon.

GewoonMaarten:
> In storybook, the single icon story does not show any icon.
> The show code does not seem correct either for single icon.

I've updated the `show code` block

DanielvanVliet:
I think it would be nice if we update IconButton to use our own icons

GewoonMaarten:
> I think it would be nice if we update IconButton to use our own icons

I've updated all stories that use material-ui/icons and replaced them with our own icons. I've added @gemeente-denhaag/icons as devDependencies. Something I would like to do with all other packages. I think this is much neater than relative imports.

DanielvanVliet (approved):

DanielvanVliet:
I don't think we need `@material-ui/icons` anymore, can you remove it?

joostvanviegen (approved):

PR #161, opened 2021-05-05
Fix/#91 group components

### LINKED ISSUE #91: Merge components into one package
Some components are published separately, but it makes no sense to use them without other components.

### Tasks
- [x] Create one package for the following components
  * Components into `tab`
    * tab
    * tabcontext
    * tablist
    * tabpanel
    * tabs
    * tabscrollbutton
  * Components into `menu`
    * menuitem
    * menulist
    * popover
    * popper
  * Components into `list`
    * listitem
    * listitemavatar
    * listitemicon
    * listitemsecondaryaction
    * listitemtext
    * listsubheader
  * Components into `avatar`
    * avatargroup
  * Components into `card`
    *  cardactions
    * cardcontent
    * cardheader
    * paper
  * Components into `accordion`
    * accordionactions
    * accordiondetails
    * accordionsummary 
- [ ] Remove the old packages

GewoonMaarten:
This issue was on the backlog for too long, so i decided to pick it up.

Packages to be removed/deprecated from NPM:
```
@gemeente-denhaag/accordionactions
@gemeente-denhaag/accordiondetails
@gemeente-denhaag/accordionsummary
@gemeente-denhaag/avatargroup
@gemeente-denhaag/cardactions
@gemeente-denhaag/cardcontent
@gemeente-denhaag/cardheader
@gemeente-denhaag/gridlisttile
@gemeente-denhaag/gridlisttilebar
@gemeente-denhaag/listitem
@gemeente-denhaag/listitemavatar
@gemeente-denhaag/listitemicon
@gemeente-denhaag/listitemsecondaryaction
@gemeente-denhaag/listitemtext
@gemeente-denhaag/listsubheader
@gemeente-denhaag/menuitem
@gemeente-denhaag/menulist
@gemeente-denhaag/mobilestepper
@gemeente-denhaag/paper
@gemeente-denhaag/popover
@gemeente-denhaag/popper
@gemeente-denhaag/radiogroup
@gemeente-denhaag/step
@gemeente-denhaag/stepbutton
@gemeente-denhaag/stepconnector
@gemeente-denhaag/stepcontent
@gemeente-denhaag/stepicon
@gemeente-denhaag/steplabel
@gemeente-denhaag/tabcontext
@gemeente-denhaag/tablist
@gemeente-denhaag/tabpanel
@gemeente-denhaag/tabscrollbutton
@gemeente-denhaag/tabs
```
These can be removed after this PR merges with development. See https://docs.npmjs.com/unpublishing-packages-from-the-registry how to remove the packages.

closes #91


DanielvanVliet (changes requested):
Can you also change each package.json to include the correct repository directory?

scripts/create-references.js
```diff
@@ -11,55 +10,42 @@ const config = JSON.parse(fs.readFileSync('tsconfig.json').toString());
...
-                }
-            }
-            fs.writeFileSync(
-                tsconfigPath,
-                JSON.stringify(workspaceConfig, undefined, 4),
-            );
+(async function () {
+  const { stdout, stderr } = await exec('yarn workspaces info --json');
+
+  const lines = stdout.split('\n');
```
ogjtech, line 16: Apparently, '\n' line separator does not work on all devices. Although I highly doubt an Atari TOS or MS DOS machine will ever visit this site, an older version of MacOS might and that only recognizes '\r' as end-of-line char.

For accessibility sake, to support all possible end-of-line characters ever, the value here should be `/\r\n|\n\r|\n|\r/g` (note, the order of the chars in this string is very important, I recommend copy-pasting it directly).

Here is the [source](https://dirask.com/posts/JavaScript-how-to-split-string-by-newline-k1wEQp).

scripts/create-references.js
```diff
@@ -11,55 +10,42 @@ const config = JSON.parse(fs.readFileSync('tsconfig.json').toString());
...
+  const depthTree = lines.slice(1, lines.length - 2).join('\n');
+  let workspaces = undefined;
+
+  // sometimes yarn prints this line: yarn run v1.22.10 and sometimes not
```
ogjtech, line 20: But does it work every time? Differing outputs per run can be confusing and is prone to causing more issues as there is no way of confirming whether the command has run properly. Perhaps find out why it has different outputs and write that in the comment as well?

scripts/create-references.js
```diff
@@ -11,55 +10,42 @@ const config = JSON.parse(fs.readFileSync('tsconfig.json').toString());
...
+  try {
+    workspaces = JSON.parse(depthTree);
+  } catch (e) {
+    workspaces = JSON.parse(stdout);
+  }
+
+  for (const name in workspaces) {
+    const workspace = workspaces[name];
+    const location = path.resolve(process.cwd(), workspace.location);
+    const tsconfigPath = path.resolve(location, 'tsconfig.json');
```
ogjtech, line 30: I would put these two variables in a `try-catch` block, as resolving paths is sensitive to errors. The path might differ on the server, builds might go wrong or (since this is JavaScript) the `path.resolve` function might just decide it's having a bad day.

In any case, you want to be able to either fallback to a back-up method, refresh the page to try again or let the user know something went wrong. This looks like a function that reaches quite deep into the inner workings of the PWA, so you want to be sure any exceptions are properly handled.

ogjtech (changes requested):

scripts/create-references.js
GewoonMaarten: `path.resolve` never throws an error, it just concatenates the path. Whether or not it exists. With `fs.existsSync(tsconfigPath)` I check if it actually exists.

This is also a developer script to automatically update the references in `tsconfig.json`, based on the dependencies of the component. The script is not really important.

scripts/create-references.js
GewoonMaarten: It adds a `\n` and fs will make sure this `\n` is properly written to the file.

scripts/create-references.js
GewoonMaarten: If I run this script manually with `node scripts/create-references.js` I get a different output compared to if I add it as a script and run it that way. 

I don't really feel like investigating this any further as its just a trivial developer script. If it doesn't work, you just need to update the references in tsconfig.json manually. It is not really a big deal.

scripts/create-references.js
ogjtech: I see that, and the `.join('\n');` function below will work as expected as it is still a valid character. However I'm not sure whether the `.split('\n');` function will always work because of the different linebreaks in some OS.

It is a niche to take this into account, but since this is a gov site I think we'd want to take into account as many possible OSes, especially as this is a dev script and some devs are infamous for working on some arbitrary OS (right, @bddjong :)).

GewoonMaarten: I've updated the split with the characters you mentioned

ogjtech (approved):

PR #169, opened 2021-05-12
Feature/#65 list design

### LINKED ISSUE #65: Design: List
The List and List related components needs to be styled according to the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1643%3A0).

## Acceptance criteria
**Components**
- [x] List
- [x] ListItem
- [x] ListItemIcon
- [x] ListItemText
- [x] ListSubHeader

- [x] ListItemSecondaryAction

**ListItem variants implemented**
- [x] Normal
- [x] With Arrow (on the right side)
- [x] With Action (on the right side)

**ListItem sizes implemented**
- [x] One line
- [x] Two lines (with subtitle `ListItemText secondary` attribute)

**ListItem states implemented**
- [x] Normal
- [x] Hover
- [x] Focus
- [x] Active

**ListItem icon support (ListItemIcon)**
- [x] Icon can provided
- [x] Icon can be placed on the left side

**Material-UI Components which we do not use**
- [x] ListItemAvatar

**ListSubHeader variants implemented**
- [x] Normal

**List variants implemented**
- [x] Normal

**Requirements**
- [x] [Alignment](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1643%3A0)  requirements are met
- [ ] Classes use the BEM naming convention

DanielvanVliet:
There is currently one accessibility violation according to the storybook plugin due to the subheader `<p>` element being placed inside the `<ul>` tag.
This appears to already be a violation in the MUI implementation and since the header is not navigable with tab or clickable with screenreader I don't think it is a serious concern.
closes #65


src/components/List/src/ListItemSecondaryAction.tsx
```diff
@@ -2,6 +2,14 @@ import React from 'react';
 import MaterialListItemSecondaryAction from '@material-ui/core/ListItemSecondaryAction';
 import { BaseChildrenProps } from '@gemeente-denhaag/baseprops';
 
+import { listitemsecondaryaction_classes as classes } from './bem-mapping';
+
+import './mui-override.module.css';
+import './listitemsecondaryaction.module.css';
+import './listitem.module.css';
+import '@gemeente-denhaag/design-tokens-components';
```
bddjong, line 10: We don't need this import here. IIRC, all tokens are up to the implementing developer to import or define in their app.

src/components/List/src/ListSubheader.tsx
```diff
@@ -1,40 +1,25 @@
...
-  disableGutters?: boolean;
-
-  /**
-   * If `true`, the List Subheader will not stick to the top during scroll.
-   */
-  disableSticky?: boolean;
+import './mui-override.module.css';
+import './listsubheader.module.css';
+import '@gemeente-denhaag/design-tokens-common';
+import '@gemeente-denhaag/design-tokens-components';
```
bddjong, line 9: Same here, no need to import these. That's for the implementer.

bddjong (changes requested):

src/components/List/src/List.tsx
```diff
@@ -1,37 +1,35 @@
...
-  dense?: boolean;
+import { StylesProvider } from '@material-ui/core';
+import { list_classes as classes } from './bem-mapping';
 
-  /**
-   * If `true`, vertical padding will be removed from the list.
-   */
-  disablePadding?: boolean;
+import './mui-override.module.css';
+import '@gemeente-denhaag/design-tokens-components';
```
bddjong, line 10: Ditto.

src/components/List/src/ListItem.tsx
```diff
@@ -1,74 +1,111 @@
...
-export interface ListItemProps extends BaseProps {
-  /**
-   * Defines the `align-items` style property.
-   */
-  alignItems?: 'flex-start' | 'center';
+import './mui-override.module.css';
+import './listitem.module.css';
+import './listitemsecondaryaction.module.css';
+import '@gemeente-denhaag/design-tokens-common';
+import '@gemeente-denhaag/design-tokens-components';
```
bddjong, line 14: Ditto.

src/components/List/src/ListItemText.tsx
```diff
@@ -1,50 +1,49 @@
...
-  /**
-   * If `true`, the children won't be wrapped by a Typography component.
-   * This can be useful to render an alternative Typography variant by wrapping
-   * the `children` (or `primary`) text, and optional `secondary` text
-   * with the Typography component.
-   */
-  disableTypography?: boolean;
+import './mui-override.module.css';
+import './listitemtext.module.css';
+import '@gemeente-denhaag/design-tokens-components';
```
bddjong, line 12: image

src/components/List/src/listitem.module.css
```diff
@@ -0,0 +1,82 @@
...
+  padding-inline: calc(var(--denhaag-listitem-padding-inline) - var(--denhaag-listitem-border-size));
+}
+
+.denhaag-list__item:active > .denhaag-list__secondaryaction {
+  right: calc(var(--denhaag-listitem-secondaryaction-right) - var(--denhaag-listitem-border-size));
+}
+
+.denhaag-list__item:focus > .denhaag-list__secondaryaction {
+  right: calc(var(--denhaag-listitem-secondaryaction-right) - var(--denhaag-listitem-border-size));
+  color: var(--denhaag-blue-3);
```
bddjong, line 27: Don't refer to proprietary styles in the component css. It's probably a good idea to create a component design token for the font color, and giving it a default value of --denhaag-blue-3.

src/components/List/src/listitem.module.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.denhaag-list__item:focus .denhaag-list__itemtextprimary {
+  color: var(--denhaag-blue-3);
```
bddjong, line 31: ditto, and a few more times in this file

src/components/List/src/listitem.module.css
```diff
@@ -0,0 +1,82 @@
...
+.denhaag-list__item[actionType="action"]:focus .denhaag-list__itemicon {
+  color: var(--denhaag-grey-4);
+}
+
+.denhaag-list__item[secondarytext] {
+  padding-block: var(--denhaag-listitem-padding-block-secondary);
+}
+
+.denhaag-list__item[secondarytext]:active {
+  padding-block: calc(var(--denhaag-listitem-padding-block-secondary) - var(--denhaag-listitem-border-size));
```
bddjong, line 69: Is this still required if we set box-sizing to border-box?

src/components/List/src/listitem.module.css
```diff
@@ -0,0 +1,82 @@
...
+.denhaag-list__item[secondarytext]:focus {
+  padding-block: calc(var(--denhaag-listitem-padding-block-secondary) - var(--denhaag-listitem-border-size));
+}
+
+.denhaag-list__item[actionType="action"] {
+  cursor: default;
+}
+
+.denhaag-list__item.denhaag-list__itemfocused {
+  background-color: white;
```
bddjong, line 81: Perhaps this should be a token

src/styles/Components/src/tokens-list.css
```diff
@@ -0,0 +1,35 @@
...
+  --denhaag-listitem-padding-block-secondary: 8px;
+
+  /* ListItemIcon */
+  --denhaag-listitem-icon-color: var(--denhaag-grey-4);
+
+  /* ListItemSecondaryAction */
+  --denhaag-listitem-secondaryaction-focus-color: var(--denhaag-blue-3);
+  --denhaag-listitem-secondaryaction-color: var(--denhaag-grey-4);
+  --denhaag-listitem-secondaryaction-right: 16px;
+  --denhaag-listitem-secondaryaction-padding: 4px;
```
bddjong, line 20: I think: `--denhaag-list-item-secondary-action` would be better

src/components/List/src/bem-mapping.ts
```diff
@@ -0,0 +1,32 @@
...
+    root: 'denhaag-list__secondaryaction'
+}
+
+export const listitemtextprimary_classes = {
+    root: 'denhaag-list__itemtextprimary'
+}
+
+export const listitemtextsecondary_classes = {
+    root: 'denhaag-list__itemtextsecondary'
+}
```
GewoonMaarten, line 32: no new line?!!1!

src/styles/Components/src/tokens-list.css
mjcarsjens: Agreed, we should always use `-` in between words where in regular language there would be a space. When in doubt use this rule: when a property/component name in MUI has a capital letter in the middle of its name this indicates the start of a new word, meaning we should insert a `-` in that position in the design token naming.

src/components/List/src/bem-mapping.ts
```diff
@@ -0,0 +1,32 @@
...
+export const listitemtext_classes = {
+  root: 'denhaag-list__itemtext',
+};
+
+export const listitemsecondaryaction_classes = {
+  root: 'denhaag-list__secondaryaction',
+};
+
+export const listitemtextprimary_classes = {
+  root: 'denhaag-list__itemtextprimary',
```
mjcarsjens, line 27: We're getting into some rather complex subcomponents here, so the BEM naming also gets a bit complicated here. 
In this case it should at least be `item-text` instead of `itemtext`. But furthermore, are these two different elements, or are these basically the same element with a modifier applied to it?

In the first case I would argue we should use `denhaag-list__item-primary-text` and `denhaag-list__item-secondary-text`. That way `primary` and `secondary` don't look like a modifier with improperly applied BEM naming and we use a sort of natural language flow and remain consistent with the naming of for example `denhaag-list__item-secondary-action`.

If we decide these are actually both just the element (for example) `denhaag-list__item-text-content` and the secondary text just has some different styling than the primary text we would get the following classes: `denhaag-list__item-text-content` and `denhaag-list__item-text-content--secondary`. Note that a modifier class is always a class you add, you do not use it to replace the original class. This would mean for our secondary text we would get `root: denhaag-list__item-text-content denhaag-list__item-text-content--secondary`. 
Another option would be to do the following:

```
export const listitemtextcontent_classes = {
   root: 'denhaag-list__item-text-content',
   body2: 'denhaag-list__item-text-content--secondary'
}
```

And apply `variant="body2"` to the Typography component, in which case the MUI component would automatically add the correct class, but this implementation is highly dependent on our implementation of the Typography component so I would at this moment stay away from this.

src/components/List/src/listitem.module.css
```diff
@@ -0,0 +1,78 @@
...
+
+.denhaag-list__item:hover .denhaag-list__itemtextprimary {
+  color: var(--denhaag-listitemtext-primarytext-hover-color);
+}
+
+.denhaag-list__item:hover .denhaag-list__itemicon {
+  color: var(--denhaag-listitem-icon-hover-color);
+}
+
+.denhaag-list__item[secondarytext] .denhaag-list__itemtextprimary {
```
mjcarsjens, line 49: Does the attribute `secondary-text` actually exist, I can't seem to find it anywhere.
Furthermore, I would prefer using a modifier in this scenario, for example `.denhaag-list__item--multi-line`

src/styles/Components/src/tokens-list.css
```diff
@@ -0,0 +1,40 @@
...
+
+  /* ListItemText */
+  --denhaag-listitemtext-primarytext-focus-color: var(--denhaag-blue-3);
+  --denhaag-listitemtext-primarytext-hover-color: var(--denhaag-blue-3);
+  --denhaag-listitemtext-primary-color: var(--denhaag-grey-4);
+  --denhaag-listitemtext-primary-weight: 400;
+  --denhaag-listitemtext-secondary-color: var(--denhaag-grey-4);
+  --denhaag-listitemtext-secondary-weight: 400;
+  --denhaag-listitemtext-secondary-font-size: var(--denhaag-font-size-s);
+  --denhaag-listitemtext-secondary-primarytext-line-height: 32px;
```
mjcarsjens, line 33: The naming here does get a bit complex, but i would prefer using the term `multi-line` instead of `secondary` here.

src/styles/Components/src/tokens-list.css
DanielvanVliet: I applied the - naming convention to all design tokens.

src/components/List/src/bem-mapping.ts
DanielvanVliet: I agree the subcomponents become a bit more complex here.
The `denhaag-list__item-text`  is applied to the react component (div) that is the parent to two distinct header elements `denhaag-list__item-text-primary` and `-secondary`, where secondary is optional.
Something similar to this:
```
<div class="denhaag-list__item-text">
  <h class="list__item-text-primary"></h>
  <h class="list__item-text-secondary"></h>
<div>
```
Since primary and secondary are not different styles applied to the same element, I don't think we should use a --flag here.

I will do some renaming though to make it more readable and consistent.

src/components/List/src/listitem.module.css
DanielvanVliet: Good find, I forgot to rename the secondaryText property to secondarytext so css did not find it.
Solved a bug 👍!

src/styles/Components/src/tokens-list.css
DanielvanVliet: It is necessary to specify secondary here since it is aimed at another element.
I agree this is confusing  since secondary was also used to specify primary text in the multiline case here.
So I renamed the secondary in the case of primarytext to multi-line and left all other secondary as they are.

src/components/List/src/List.stories.tsx
```diff
@@ -1,19 +1,18 @@
...
+import { action } from '@storybook/addon-actions';
+
+import './mui-override.module.css';
+import './listitem.module.css';
+import './listitemsecondaryaction.module.css';
+import './listitemicon.module.css';
+import './listitemtext.module.css';
+import './listsubheader.module.css';
+import '@gemeente-denhaag/design-tokens-common';
+import '@gemeente-denhaag/design-tokens-components';
```
bddjong, line 15: We shouldn't have to import the token

src/components/List/src/ListItem.tsx
```diff
@@ -1,74 +1,110 @@
...
    */
-  component?: React.ElementType<React.HTMLAttributes<HTMLElement>>;
+  actionType?: 'nav' | 'action';
 
   /**
-   * The container component used when a `ListItemSecondaryAction` is the last child.
+   * The callback fired when clicked
    */
-  ContainerComponent?: React.ElementType<React.HTMLAttributes<HTMLDivElement>>;
+  onClick?: (event: React.MouseEvent) => void;
```
bddjong, line 26: I think we want to type this the following way to support touch events: `event: React.MouseEvent | React.TouchEvent`. Check out the ButtonProps for reference.

bddjong (changes requested):

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
bddjong, line 1: newline

src/components/List/src/List.tsx
```diff
@@ -1,37 +1,31 @@
...
-  subheader?: React.ReactElement;
+  subheader: React.ReactElement<ListSubheaderProps>;
 }
 
 /**
  * Primary UI component for user interaction
  */
 export const List: React.FC<ListProps> = (props: ListProps) => {
-  return <MaterialList {...props}>{props.children}</MaterialList>;
+  const materialProps = { ...props, subheader: props.subheader };
```
bddjong, line 20: TODO: remove `...props` to explicit properties.

src/components/List/src/listitem.css
```diff
@@ -0,0 +1,86 @@
+.denhaag-list__item {
+  padding-block-start: var(--denhaag-list-item-padding-block);
+  padding-block-end: var(--denhaag-list-item-padding-block);
+  padding-inline-start: var(--denhaag-list-item-padding-inline);
+  padding-inline-end: var(--denhaag-list-item-padding-inline);
+}
+
+.denhaag-list__item:focus {
```
mjcarsjens, line 8: For the other items we added a modifier rule for all CSS states (`hover`, `focus`, `active`): `.denhaag-list__item--focus`.
This is done for easier visual regression testing which at this point we're not actually doing, but I still would like to see it added just in case.

mjcarsjens (commented):
Nice work! Not much to add at this point, other than that it seems like when I run it locally there seems to be something wrong with the styling.
image

bddjong:
> Nice work! Not much to add at this point, other than that it seems like when I run it locally there seems to be something wrong with the styling.

That styling is looking fine on my machine, I think it's a local issue for you.

bddjong (approved):

src/components/List/src/listitem.css
DanielvanVliet: I've added this for the easy cases, others were not as easy.
For instance what rule should be used here:

`.denhaag-list__item[secondarytext]:active`

I've added an issue to resolve this in the future, and I think we can merge this PR.

bddjong (approved):

PR #172, opened 2021-05-15
Implemented design for Divider component.

ogjtech:
The design tokens and their values are defined as follows:

```css
:root {
    --denhaag-divider-border-color: var(--denhaag-grey-2);
    --denhaag-divider-border-width: 1px;
    --denhaag-divider-margin: 16px 16px;
}
```

The margin property was defined as above, since the two variants (vertical and horizontal) have their margins set using `margin-block` and `margin-inline` respectively. Both these CSS keywords require two values (above and below for horizontal, left and right for vertical) and since both variants need the same amount of margin based on the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1596%3A83), it would be redundant to define separate values for this.


src/components/Divider/divider.module.css
```diff
@@ -0,0 +1,17 @@
...
+  --denhaag-divider-margin: 16px 16px;
+}
+
+.denhaag-divider {
+  height: var(--denhaag-divider-border-width);
+  background: var(--denhaag-divider-border-color);
+  margin-block: var(--denhaag-divider-margin);
+}
+
+.denhaag-divider-vertical {
```
Robbert, line 13: Following the BEM syntax should be `denhaag-divider--vertical`, with an additional dash.

src/components/Divider/divider.module.css
```diff
@@ -0,0 +1,17 @@
+:root {
+  --denhaag-divider-border-color: var(--denhaag-grey-2);
+  --denhaag-divider-border-width: 1px;
+  --denhaag-divider-margin: 16px 16px;
+}
```
bddjong, line 5: These should be moved to the basestyles package. Look at the button PR. Creating a file called `tokens-divider.css` and using an `@import "tokens-divider.css"` in `globals.module.css`. It's unlikely that that's the definitive approach, but at the very least it groups our tokens in a single package, making it easier to refactor later.

src/components/Divider/divider.module.css
```diff
@@ -0,0 +1,17 @@
+:root {
+  --denhaag-divider-border-color: var(--denhaag-grey-2);
+  --denhaag-divider-border-width: 1px;
+  --denhaag-divider-margin: 16px 16px;
```
Robbert, line 4: While in The Hague designs they margins just so happen to have the same values, this is likely not the case for other designs. I would propose to have the `margin-inline` and `margin-block` separation already in place for that.

bddjong (changes requested):

src/components/Divider/src/Divider.stories.tsx
```diff
@@ -18,98 +19,58 @@ export default {
...
+    </Grid>
+  ) : (
+    <List>
+      <ListItem button>
+        <ListItemIcon>
+          <InboxIcon />
+        </ListItemIcon>
+        <ListItemText primary="I am above the divider." />
+      </ListItem>
+      <Divider {...args} />
```
Robbert, line 37: I know the `<Divider/>` is used for this purpose in Material-UI, but is this really something to strive for? Couldn't this be implemented as `list--dividers` or something like that with `border-bottom`? I feel that ideally the Divider component has `role="separator"` by default, but that has a certain meaning and effect for screen readers. For list items that would not be appropriate, whereas for separating texts or separating toolbar buttons into groups it would be appropriate.

Robbert:
Is there already a use case for the vertical divider? Perhaps it could be possible to postpone including it in the design system if there is no actual use for it.

src/components/Divider/src/Divider.stories.tsx
bddjong: Perhaps it's a better idea to split them into different components as you mentioned earlier. Our use case for the divider is to split visual elements, not necessarily a screen reader separator.

Robbert:
It would be nice if you could include the documentation from Figma in a `.md` file that is displayed in Storybook. A while back we discussed with Roos that migrating the docs to Storybook and have that as single source of truth could be nice.

It would also be very convenient to also include the actual Figma frame in Storybook with an iframe:

```jsx
<MDXEmbedProvider>
  <Figma title="Backlink" url="file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1596%3A83" />
</MDXEmbedProvider>
```

bddjong:
> It would be nice if you could include the documentation from Figma in a `.md` file that is displayed in Storybook. A while back we discussed with Roos that migrating the docs to Storybook and have that as single source of truth could be nice.
> 
> It would also be very convenient to also include the actual Figma frame in Storybook with an iframe:
> 
> ```js
> <MDXEmbedProvider>
>   <Figma title="Backlink" url="file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1596%3A83" />
> </MDXEmbedProvider>
> ```

The addition of design system documentation in Storybook is currently being headed by Roos and @mjcarsjens . I think it would be nice for the RYSST development team if we could keep that out of our scope for these issues

src/components/Divider/src/Divider.stories.tsx
Robbert: If this Divider element is for the presentational variant, I don't think it should be implemented with an `<hr>` like in Material-UI, because that has `role="separator"` implied. Also their suggestion to use the inset variant that renders as `<li role="separator">` doesn't seem very accessible. It feels to me that it would only be appropriate to separate groups of list items, not to separate every list item.

Also the story of separating text content like `<p>...</p><Divider/><p>...</p>` should not be included in the examples for a visual divider, since those should have the "semantic" separator.

Anyway, I agree there should be two separate components in that case. I think you would need to implement the visual divider yourself and not rely on Material-UI in this case.

ogjtech: The vertical variant separates two `<p>` elements in a MUI `<Grid>` component, separating text. Whilst this Divider component is not the separator component you are referring to, I do think that it does also serve that purpose. In the background this component is just an hr, so I think applying role="separator" to this Divider may result in the same accessibility-wise.

ogjtech:
> 
> 
> Is there already a use case for the vertical divider? Perhaps it could be possible to postpone including it in the design system if there is no actual use for it.

@Robbert there does not appear to be one in the demonstrated designs in Figma, but it is defined in the variants on the same Figma. I cannot 100% recall, but I think I do remember seeing examples of the vertical divider in other designs. I think it better to include it for now and remove it afterwards. Better to have more options to start with and narrow it down as we progress.

Robbert:
@ogjtech Fine with me 👍

My perspective on adding vs. removing is from a semantic versioning point of view: adding is a _minor_ release to which everyone can safely upgrade, while removing is a breaking change and a _major_ release to which you cannot automatically upgrade. Major releases often cause people to stay stuck on older versions, so they are best avoided, or else you might need to provide security patches to many old releases too because they are still frequently used. That's why I would rather do a gradual release of features and avoid any breaking changes if I can.

bddjong:
> @ogjtech Fine with me 
> 
> My perspective on adding vs. removing is from a semantic versioning point of view: adding is a _minor_ release to which everyone can safely upgrade, while removing is a breaking change and a _major_ release to which you cannot automatically upgrade. Major releases often cause people to stay stuck on older versions, so they are best avoided, or else you might need to provide security patches to many old releases too because they are still frequently used. That's why I would rather do a gradual release of features and avoid any breaking changes if I can.

@Robbert @ogjtech I agree we should avoid adding variants if there's a risk we have to remove them at a later stage. We can always easily add it if people actually request it. For now I've asked if there's a use case for the vertical divider. Lets keep the PR open till the designers confirm it.

src/components/Divider/src/Divider.stories.tsx
ogjtech: > Also the story of separating text content like `<p>...</p><Divider/><p>...</p>` should not be included in the examples for a visual divider, since those should have the "semantic" separator.

The purpose of this example wasn't so much to split text, it was moreso to demo the ability to vertically divide two components. Will push a commit with different components in a bit.

ogjtech:
@Robbert @bddjong while I do agree its safer to add something afterwards, the library is still in development (we are live, but we do not have a major release yet). Material UI has their release policy defined on their [website](https://material-ui.com/versions/#deprecation-practices).

Seasoned developers are not likely to install a library which does not have any major releases, maybe to demonstrate it or to keep up with its progress. Knowing this I do not think we should press too hard on "breaking changes" just yet, especially since we're going to introduce many when it comes to the design tokens.

src/styles/BaseStyles/src/tokens-divider.css
```diff
@@ -0,0 +1,8 @@
+:root {
+  --denhaag-divider-border-color: var(--denhaag-grey-2);
+  --denhaag-divider-border-width: 1px;
+  --denhaag-divider-margin-block-start: 16px;
+  --denhaag-divider-margin-block-end: 16px;
+  --denhaag-divider-margin-inline-start: 16px;
+  --denhaag-divider-margin-inline-end: 16px;
+}
```
Robbert, line 8: Design tokens are looking good, great work 👍

ogjtech:
@Robbert @bddjong apparently, an `<hr />` HTML5 tag (the underlying component for the MUI Divider) implicitly has a separator role. [MUI advises](https://material-ui.com/components/dividers/#html5-specification) to, when a Divider is used in a List or Menu-type component, specify the underlying component to be `<li />` instead of `<hr />` in order to match the HTML5 specification.

Specifying it as a `<li />`-component by default seems bad practice to me, as we cannot assume it will be exclusively used in lists. However, I am unsure which substitute component would best replace this component... Any ideas?

mjcarsjens:
> 
> 
> @Robbert @bddjong apparently, an `<hr />` HTML5 tag (the underlying component for the MUI Divider) implicitly has a separator role. [MUI advises](https://material-ui.com/components/dividers/#html5-specification) to, when a Divider is used in a List or Menu-type component, specify the underlying component to be `<li />` instead of `<hr />` in order to match the HTML5 specification.
> 
> Specifying it as a `<li />`-component by default seems bad practice to me, as we cannot assume it will be exclusively used in lists. However, I am unsure which substitute component would best replace this component... Any ideas?

adding `role="presentation"` will ensure that it is ignored by for example screenreaders.

ogjtech:
> adding `role="presentation"` will ensure that it is ignored by for example screenreaders.

@mjcarsjens I just now added `role={"presentation"}` to the component. I am yet to test whether it is actually carried over to the underlying `<hr \>` element in the MUI component, but will do soon as I figure out why Storybook is refusing to start up _again_.

ogjtech:
@mjcarsjens @Robbert the `role={'presentation'}`-tag carries over to the underlying `<hr \>` element. Will resolve conflicts and then re-request reviews 👍🏻

src/components/Divider/divider.module.css
```diff
@@ -0,0 +1,20 @@
...
+  margin-block-start: var(--denhaag-divider-margin-block-start);
+  margin-block-end: var(--denhaag-divider-margin-block-end);
+}
+
+.denhaag-divider--vertical {
+  margin-inline-start: var(--denhaag-divider-margin-inline-start);
+  margin-inline-end: var(--denhaag-divider-margin-inline-end);
+  width: var(--denhaag-divider-border-width);
+  height: inherit;
+}
```
bddjong, line 20: This file should be in the  src folder of the Divider component.

bddjong (changes requested):

src/components/Divider/divider.module.css
```diff
@@ -0,0 +1,20 @@
+:root {
+  --denhaag-divider-border-color: var(--denhaag-grey-2);
+  --denhaag-divider-border-width: 1px;
+  --denhaag-divider-margin: 16px 16px;
+  --denhaag-divider-vertical-margin-inline: 16px 16px;
```
bddjong, line 5: The tokens should only be defined in the `styles/components/src/tokens-divider.css` file (with a few exceptions). Some are double now, exisiting both in the component tokens file and in the component css here.

src/styles/Components/src/tokens-divider.css
```diff
@@ -0,0 +1,8 @@
+:root {
+  --denhaag-divider-border-color: var(--denhaag-grey-2);
+  --denhaag-divider-border-width: 1px;
+  --denhaag-divider-margin-block-start: 16px;
+  --denhaag-divider-margin-block-end: 16px;
+  --denhaag-divider-margin-inline-start: 16px;
+  --denhaag-divider-margin-inline-end: 16px;
```
bddjong, line 7: Perhaps it would be nice to add aggregation tokens for `-block` and `-inline`, since by default they're all 16px. We would still have the `-start` and `-end` specific tokens, so users _can_ overwrite them if they want, but have them default to a `denhaag-divider-margin-{block | inline}: 16px` token.

The current tokens would look like this:

```css
--denhaag-divider-margin-block-start: var(--denhaag-divider-margin-block); 
--denhaag-divider-margin-block-end: var(--denhaag-divider-margin-block); 
--denhaag-divider-margin-inline-start: var(--denhaag-divider-margin-inline); 
--denhaag-divider-margin-inline-end: var(--denhaag-divider-margin-inline);
```

@Robbert @mjcarsjens please correct me if it's a rubbish idea :P

src/components/Divider/src/index.tsx
```diff
@@ -35,7 +17,16 @@ export interface DividerProps extends BaseDataDisplayClassesProps {
...
 export const Divider: React.FC<DividerProps> = (props: DividerProps) => {
-  return <MaterialDivider {...props} />;
+  const classes = {
+    root: 'denhaag-divider',
+    vertical: 'denhaag-divider--vertical'
+  }
+
+  return (
+    <StylesProvider injectFirst>
+      <MaterialDivider variant={"fullWidth"} classes={classes} role={"presentation"} {...props} />
```
bddjong, line 27: We're not using `...props` anymore. It risks exposing MaterialUI API we don't want to support. Pass every prop in our interface explicitly.

src/styles/Components/src/tokens-divider.css
ogjtech: I mean, personally I do not think defining aggregation token for default margins adds any value here. Also, in your example, if someone decides they want `-inline-start` to be 32px, they may get confused seeing yet another design token.

bddjong: As a developer, I'd rather define just two tokens (the aggregated tokens) than 4 tokens (start, end for both block and inline). Especially if they're all the same, which is fairly likely.

ogjtech: I am unsure whether or not this adds value. I mean, a dev could just as easily define his own aggregated tokens and insert those, that makes it easier for them to maintain rather than digging through our tokens. In my opinion this is a sort of "less is more" scenario. The aggregated tokens may help in more clearly defining things in our library, but leaving it this way I think leaves more room for customization. Curious what @Robbert and @mjcarsjens think about this..

src/components/Divider/src/index.tsx
```diff
@@ -1,32 +1,14 @@
 import React from 'react';
 import MaterialDivider from '@material-ui/core/Divider';
 import { BaseDataDisplayClassesProps } from '@gemeente-denhaag/basedatadisplayprops';
+import { StylesProvider } from '@material-ui/core';
+import '../divider.module.css';
```
bddjong, line 5: Shouldn't this be ./divider.module.css

GewoonMaarten (approved):

bddjong (approved):

PR #189, opened 2021-05-25
Added onClick prop to IconButton

DanielvanVliet:
Quick addition to IconButton, I wonder how this was missed as it seems unusable without this prop.

### DIFF
--- a/src/components/IconButton/src/index.tsx
+++ b/src/components/IconButton/src/index.tsx
@@ -35,6 +35,11 @@ export interface IconButtonProps extends BaseProps {
    * `small` is equivalent to the dense button styling.
    */
   size?: 'small' | 'medium';
+
+  /**
+   * Callback fired when the IconButton is clicked.
+   */
+  onClick?: (event: React.ChangeEvent<unknown>) => void;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/IconButton/src/index.tsx
```diff
@@ -35,6 +35,11 @@ export interface IconButtonProps extends BaseProps {
    * `small` is equivalent to the dense button styling.
    */
   size?: 'small' | 'medium';
+
+  /**
+   * Callback fired when the IconButton is clicked.
+   */
+  onClick?: (event: React.ChangeEvent<unknown>) => void;
```
bddjong, line 42: Shouldn't the type be a `React.MouseEvent` or something similar

DanielvanVliet: Yep you're right that's more precise.

bddjong (approved):

GewoonMaarten (approved):

GewoonMaarten:
You can also use `onClick?: React.MouseEventHandler<HTMLButtonElement>`

bddjong:
> You can also use `onClick?: React.MouseEventHandler<HTMLButtonElement>;`

We decided on MouseEvent since MouseEventHandler is just an alias which in our opinion doesn't add much

DanielvanVliet:
The `(event: MouseEvent<..>) => void` communicates the usage better / the user doesn't need to lookup what a handler is.

PR #191, opened 2021-05-25
Added the tabindex property to IconButton

DanielvanVliet:
This might not be something we want to expose like this, but tabIndex is necessary for the design implementation for list.

### DIFF
--- a/src/components/IconButton/src/index.tsx
+++ b/src/components/IconButton/src/index.tsx
@@ -35,11 +35,6 @@ export interface IconButtonProps extends BaseProps {
    * `small` is equivalent to the dense button styling.
    */
   size?: 'small' | 'medium';
-
-  /**
-   * The tab-index of the IconButton.
-   */
-  tabIndex?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/components/IconButton/src/index.tsx
```diff
@@ -35,6 +35,11 @@ export interface IconButtonProps extends BaseProps {
    * `small` is equivalent to the dense button styling.
    */
   size?: 'small' | 'medium';
+
+  /**
+   * The tab-index of the IconButton.
+   */
+  tabIndex?: number;
```
bddjong, line 42: I'm fine with exposing this property. Users should be able to decide for themselves how to use a component under their own circumstances.

bddjong (approved):

GewoonMaarten (dismissed):

src/components/IconButton/src/index.tsx
bddjong: Move this to baseprops

bddjong (changes requested):
Submit rereview when tabindex has been moved to baseprops

joostvanviegen (approved):

GewoonMaarten (approved):

PR #193, opened 2021-05-26
Reproducible builds using exact version from Yarn lockfile

Robbert:
I noticed we had a lot of issues with situations that were not easy to reproduce across machines and a lot of "works on my machine" situations. This PR changes the yarn installation command to use the exact versions that were tested in the PR that updated the dependencies.

Since this repository uses Dependabot to stay up to date with the latest versions of dependencies (and incidentally the test scripts of GitHub Actions will test if the upgrade would break anything) there should be no need to rely on the flexible version ranges (`^1.0.0` and `~1.0.0`) in the `package.json` file.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -25,7 +25,7 @@ jobs:
       - name: yarn install, build storybook
         run: |
           npm install -g yarn
-          yarn install
+          yarn install --frozen-lockfile
           yarn build-storybook
 
       - name: Upload storybook
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname $0)/_/husky.sh"
 
-yarn install
+yarn install --frozen-lockfile
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ class Example extends Component {
 4. Run the following command to install all development dependencies:
 
 ```bash
-yarn install
+yarn install --frozen-lockfile
 ```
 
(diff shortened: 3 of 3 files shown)


joostvanviegen (dismissed):

joostvanviegen (approved):

mjcarsjens:
@Robbert I definitely feel like this would help a lot. How would you feel about instead creating a `.yarnrc` file with `--install.frozen-lockfile true` in it? I feel like this would ensure people actually use it correctly. Not sure whether this will break dependabot though, depends on how it works internally. I know this was an issue with renovatebot where it would simply run a `yarn install` and then commit the new lockfile and create a PR, however when using a `.yarnrc` as specified above this would not update the `yarn.lock`.

mjcarsjens (approved):
I approve these changes and agree with this addition. Quite curious about the use of `.yarnrc` instead but this might break dependabot.

Robbert:
@mjcarsjens I agree with the `.yarnrc` approach, but since I had some bad experiences with solving this with `.npmrc` with both Dependabot and Snyk, I didn't attempt to do this now.

mjcarsjens:
> 
> 
> @mjcarsjens I agree with the `.yarnrc` approach, but since I had some bad experiences with solving this with `.npmrc` with both Dependabot and Snyk, I didn't attempt to do this now.

That's fair, understandable! @bddjong in my opinion we can merge this PR, might want to experiment at a later moment with the `.yarnrc` approach but for now it is still unsure whether or not this will break Dependabot.

PR #204, opened 2021-05-29
Add type to onClick function in ButtonProps

### LINKED ISSUE #103: Button onClick props not consistent
Currently the onClick props of Button like this:
`onClick?: () => void;`
While MenuItem's onClick props look as follows:
`onClick?: (event: React.ChangeEvent<unknown>) => void;`

I suggest we add the event (possibly as optional '?') to the button onClick props.

**Tasks**
- [x] Button onClick props updated
- [x] Storybook updated

bddjong:
The onClick function in the ButtonProps interface has no type. React provides [Synthetic events](https://reactjs.org/docs/events.html) that wrap around native browser events. This PR allows users to access a MouseEvent.

**Closing issues**
closes #103 Button OnClick prop type is incorrect

### DIFF
--- a/src/components/Button/src/index.tsx
+++ b/src/components/Button/src/index.tsx
@@ -10,7 +10,7 @@ export interface ButtonProps extends BaseProps {
   /**
    * Simple click handler
    */
-  onClick?: () => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
 
   /**
    * Size of the component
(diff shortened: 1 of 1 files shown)


GewoonMaarten (approved):

DanielvanVliet (approved):

src/components/Button/src/index.tsx
```diff
@@ -10,7 +10,7 @@ export interface ButtonProps extends BaseProps {
   /**
    * Simple click handler
    */
-  onClick?: () => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
```
mjcarsjens, line 13: Does this work for touchscreens though? React.TouchEvent is also a typing, which I think will be passed to the onClick when touched instead of clicked, resulting in a typing error.

mjcarsjens (commented):
Touchscreens?

src/components/Button/src/index.tsx
bddjong: Hmmm good point. I hadn't considered that. Will have a look into that

PR #206, opened 2021-05-29
Introduced disallow any rule

### LINKED ISSUE #152: Disallow use of any type
We configure eslint to trow an error if we use the `any` type. If we allow ourselves to use `any`, then there is really no reason to use typescript. Currently it throws a warning, but this should be changed to error.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md

### Tasks
- [x] Configure typescript to throw error on use (`"@typescript-eslint/no-explicit-any": "error"`)
- [x] Remove all uses of `any` (36 occurrences)

DanielvanVliet:
**Closing issues**
closes #152 

Enabled any error in eslint, and fixed/reworked all cases where any was used.

In most cases the MUI type could be copied 1:1.
In other cases the most commonly used types were chosen (e.g. value for tab).

If there was no obviously commonly used type and MUI uses any in their docs,
 I added a disable warning comment and left it as any.

In some cases the type was too complex or deeply nested to figure out (e.g. Popover), 
in that case I chose to use the type used in the Material properties.
This should work in the current state, and once we refactor these components we can revisit them.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -106,6 +106,7 @@
     "no-useless-concat": "error",
     "no-useless-escape": "error",
     "no-useless-return": "error",
+    "@typescript-eslint/no-explicit-any": "error",
     "no-void": "error",
     "no-with": "error",
     "prefer-regex-literals": "error",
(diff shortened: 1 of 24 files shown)


GewoonMaarten (approved):

bddjong:
Try to add the issue number in the branch name :D, this would be `fix/#152-disallow-any`

bddjong (approved):

PR #212, opened 2021-06-02
Fix/storybook typecheck step

### LINKED ISSUE #162: CI Storybook typescript typecheck step
We need a typescript typecheck step for storybook files in order to break the CI when a storybook file is incorrect.

The [Storybook Typescript documentation](https://storybook.js.org/docs/react/configure/typescript#mainjs-configuration) describes how to enable type checking.

### Tasks
- [ ] enable type checking in the storybook config
- [ ] enable eslint for storybook (also described in the linked doc)
- [ ] use correct tsconfig.json (the config where storybook resides)
- [ ] figured out the correct value for the `async` setting. It maybe worth while to set this to `false`.
- [ ] check if any other settings need to be enabled.

DanielvanVliet:
This is a different approach than was described in the issue, since I had trouble getting that to work on our multi-component project.

Added a typescript tsconfig file that scans all *.stories.tsx files and typechecks them.
As a result of this I had to fix type-errors and create new issues to solve problems in *.stories.tsx.

One notable change is how the button property in ListItem was changed back to any, this will be solved in a future PR #169 when list api is updated.

**Closing issues**

closes #162


src/components/FormControlLabel/src/FormControlLabel.stories.tsx
```diff
@@ -19,7 +19,7 @@ export default {
 } as Meta;
 
 const Template: Story<FormControlLabelProps> = (args: FormControlLabelProps) => (
-  <FormControlLabel label="Checkbox" control={<Checkbox {...args} />} {...args} />
+  <FormControlLabel {...args} label="Checkbox" control={<Checkbox {...args} />} />
```
GewoonMaarten, line 22: Why was this ordering changed?

src/components/Select/src/Select.stories.tsx
```diff
@@ -34,6 +34,7 @@ export function DefaultSelect(props: SelectProps): React.ReactElement {
 
   return (
     <FormControl className={classes.formControl}>
+      {/*  */}
```
GewoonMaarten, line 37: Is this comment necessary?

GewoonMaarten (changes requested):

src/components/FormControlLabel/src/FormControlLabel.stories.tsx
DanielvanVliet: If left as before, typescript considers label and control to be specified twice (once in ...args) and gives an error.
Swapping them like this causes label and control to 'override' the one in {...args} and no error occurs.

src/components/Select/src/Select.stories.tsx
DanielvanVliet: nope fixed

GewoonMaarten (approved):

mjcarsjens (changes requested):
Solid addition, let's not merge this until the issues which occur right now are fixed, but afterwards fine by me!

src/components/List/src/ListItem.tsx
```diff
@@ -19,7 +19,8 @@ export interface ListItemProps extends BaseProps {
    * If `true`, the list item will be a button (using `ButtonBase`). Props intended
    * for `ButtonBase` can then be applied to `ListItem`.
    */
-  button?: false;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  button?: any;
```
GewoonMaarten, line 23: This sucks, but I assume their is no other type that fits?

GewoonMaarten (dismissed):

src/components/List/src/ListItem.tsx
DanielvanVliet: It is a temporary fix, since when list-design is merged this property won't  be exposed anymore.

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
bddjong, line 1: no newline

src/components/Tab/src/Tab.stories.tsx
```diff
@@ -23,11 +23,11 @@ const Template: Story<TabProps> = (args: TabProps) => {
     <TabContext value={value.toString()}>
       <Tabs
         value={value}
-        onChange={(event: React.ChangeEvent<unknown>, newValue: number) => {
+        onChange={(_event: React.ChangeEvent<unknown>, newValue: number) => {
```
bddjong, line 26: Why was this renamed?

DanielvanVliet: To fix a typecheck error:
err
Eslint provided a quickfix suggesting this solution, to signal to other developers that the parameter is unused in the function.

GewoonMaarten (dismissed):

mjcarsjens (approved):

PR #221, opened 2021-06-02
Feature/#110 stepper design

### LINKED ISSUE #110: Design: Stepper
The stepper component should have a design identical to the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1616%3A6543)

## Acceptance criteria
**Components**
- [x] Stepper
- [x] Step
- [x] StepLabel
- [x] StepContent

**States**
- [x] Normal
- [x] Extended
- [x] Disabled
- [x] Hover
- [x] Focus

**Statuses/variants**
- [x] Inactive
- [x] Active
- [x] Completed
image

**Requirements**
- [x] Completed Icon should be customizable through a design token
- [x] Active and inactive colors should be customizable through design tokens
- [x] Description can be given to steps, which can be expanded
- [x] Inactive state description cannot be expanded
- [x] [Alignment](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1616%3A6543) requirements are met

GewoonMaarten:
I've renamed stepper to timeline, to match the naming in the figma design. This means that `@gemeente-denhaag/stepper` needs to be removed/marked as deprecated in the NPM registry.

**Closing issues**
closes #110


DanielvanVliet (changes requested):
- The collapse icon in the step is not completely centered.

Where is the disabled state?  😉

src/components/Timeline/src/styles/step.css
```diff
@@ -0,0 +1,15 @@
+.denhaag-timeline__step {
+  padding-block-start: 12px;
+  padding-block-end: 12px;
+}
```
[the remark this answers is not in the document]
GewoonMaarten: Added

GewoonMaarten:
I don't think I can get `<details>` to work, because not every step has a description. Maybe `role='button' aria-expanded='true/false'` could work better.

src/components/Timeline/src/StepContent.tsx
```diff
@@ -35,7 +37,13 @@ export interface StepContentProps extends BaseProps {
  * @constructor Constructs an instance of StepContent.
  */
 export const StepContent: React.FC<StepContentProps> = (props: StepContentProps) => {
-  return <MaterialStepContent {...props}>{props.children}</MaterialStepContent>;
+  return (
+    <StylesProvider injectFirst>
+      <MaterialStepContent className="denhaag-timeline__step-content" {...props}>
```
mjcarsjens, line 42: For consistencies sake I would prefer this className to also be in the bem-mapping file

src/components/Timeline/src/Step.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <MaterialStep
+      active={active}
+      classes={stepClasses}
+      expanded={isExpanded}
+      onClick={toggle}
+      onKeyDown={handleKeyDown}
+      tabIndex={tabIndex}
+      {...props}
```
mjcarsjens, line 75: I believe our general way of working is to not include `{...props}` and instead individually call every prop to disallow the end-user to adjust props of the Material-UI component which we do not want to be adjustable.

src/components/Timeline/src/StepContent.tsx
mjcarsjens: And same comment with regards to `{...props}` as in `Step.tsx`

src/components/Timeline/src/StepLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+  } else {
+    StepIconComponent = props.StepIconComponent;
+    StepIconComponent.props = { ...StepIconComponent.props, ...props.StepIconProps };
+  }
+
+  const classes = clsx('denhaag-timeline__step-label', {
+    'denhaag-timeline__step-label--active': props.active || props.expanded,
+    'denhaag-timeline__step-label--disabled': props.disabled,
+    'denhaag-timeline__step-label--completed': props.completed,
+  });
```
mjcarsjens, line 50: Nice use of clsx!

src/components/Timeline/src/StepLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if (props.StepIconComponent === undefined) {
+    StepIconComponent = (
+      <StepIcon
+        active={props.active}
+        completed={props.completed}
+        disabled={props.disabled}
+        icon={props.icon}
+        {...props.StepIconProps}
+      />
+    );
```
mjcarsjens, line 40: Generally for this use-case there is the React `defaultProps` setup, but there are some troubles with making `defaultProps` work with TypeScript afaik so not sure whether this is actually the preferred solution or not. Might be something to look into.

src/components/Timeline/src/StepIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+  } else if (props.active) {
+    color = 'action';
+  } else if (props.completed) {
+    color = 'secondary';
+  }
+
+  const textClasses = clsx('denhaag-timeline__step-icon__text', {
+    'denhaag-timeline__step-icon__text--active': props.active,
+    'denhaag-timeline__step-icon__text--completed': props.completed,
+  });
```
mjcarsjens, line 33: You can't chain together multiple "Elements" in BEM. It is always one block, followed by at most 1 element and 1 modifier. In this case this should be: `denhaag-timeline__step-icon-text`, `denhaag-timeline__step-icon-text--active`, etc. If you feel like this results in unreadable class names: it is allowed to use single underscores in the name of the element and modifier (NOT in the block name though!). So we could consider `denhaag-timeline__step-icon_text`.

src/components/Timeline/src/StepLabel.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const classes = clsx('denhaag-timeline__step-label', {
+    'denhaag-timeline__step-label--active': props.active || props.expanded,
+    'denhaag-timeline__step-label--disabled': props.disabled,
+    'denhaag-timeline__step-label--completed': props.completed,
+  });
+
+  return (
+    <div className={classes}>
+      {StepIconComponent}
+      <p className="denhaag-timeline__step-label__text">{props.children}</p>
```
mjcarsjens, line 55: Same as above with regards to chaining together multiple BEM elements in the class name.

src/components/Timeline/src/Timeline.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * @constructor Constructs an instance of Timeline.
+ */
+export const Timeline: React.FC<TimelineProps> = (props: TimelineProps) => {
+  return (
+    <MaterialStepper
+      nonLinear
+      orientation="vertical"
+      connector={<React.Fragment />}
+      classes={timelineClasses}
+      {...props}
```
mjcarsjens, line 31: Same as above wrt `{...props} `

src/components/Timeline/src/StepLabel.tsx
mjcarsjens: Scratch that, mb, this is a native HTML element instead of a Material-UI component, so might be okay to leave it here.

src/components/Timeline/src/bem-mapping.ts
```diff
@@ -0,0 +1,28 @@
...
+export const stepIconClasses = {
+  root: 'denhaag-timeline__step-icon',
+  colorAction: 'denhaag-timeline__step-icon--active',
+  colorSecondary: 'denhaag-timeline__step-icon--completed',
+  colorPrimary: 'denhaag-timeline__step-icon--inactive',
+  colorDisabled: 'denhaag-timeline__step-icon--disabled',
+};
+
+export const stepCollapseIconClasses = {
+  root: 'denhaag-timeline__step__collapse-icon',
```
mjcarsjens, line 27: `denhaag-timeline__step-collapse-icon`, or `denhaag-timeline__step_collapse-icon` (up to you which you prefer, try to be consistent in whichever solution you go for here and might be good to discuss this with the team)

src/components/Timeline/src/Timeline.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import MaterialStepper from '@material-ui/core/Stepper';
+import BaseProps from '@gemeente-denhaag/baseprops';
+
+import { timelineClasses } from './bem-mapping';
+
+export interface TimelineProps extends BaseProps {
```
ogjtech, line 7: I think we should prepend interface names with `I` for clarity. So this interface would become `ITimelineProps`. That way it is much easier to distinguish what's what and prevents devs from expecting a component where there is none.

src/components/Timeline/src/bem-mapping.ts
```diff
@@ -0,0 +1,28 @@
...
+  completed: 'denhaag-timeline__step--completed',
+};
+
+export const stepIconClasses = {
+  root: 'denhaag-timeline__step-icon',
+  colorAction: 'denhaag-timeline__step-icon--active',
+  colorSecondary: 'denhaag-timeline__step-icon--completed',
+  colorPrimary: 'denhaag-timeline__step-icon--inactive',
+  colorDisabled: 'denhaag-timeline__step-icon--disabled',
+};
```
ogjtech, line 24: Why are these classes defined here in `Timeline`, when they only have impact on the icon shown. This class name is very confusing to me, as in I know what it belongs to but I have zero indication of what this class does or what it changes. In my opinion, either replace these to the `Step` component and rename them to `step-icon--disabled` or make them more descriptive of their impact to the `Timeline` component.

src/components/Timeline/src/styles/step-content.css
```diff
@@ -0,0 +1,8 @@
+.denhaag-timeline__step-content:not(:empty) {
+  margin-block-start: var(--denhaag-timeline-step-content-margin);
+  color: var(--denhaag-timeline-step-content-color);
+}
+
+.denhaag-timeline__step-content:empty {
+  margin-block-start: 0;
```
ogjtech, line 7: This styling looks like it should belong in the MUI overrides of this class.

src/components/Timeline/src/styles/step-label.css
```diff
@@ -0,0 +1,36 @@
...
+  margin-block-end: 0;
+  margin-inline-start: 0;
+  margin-inline-end: 0;
+}
+
+.denhaag-timeline__step-label__text:hover {
+  overflow: visible;
+  white-space: normal;
+  height: auto;
+}
```
ogjtech, line 22: These look like they should be in an MUI overrides file.

src/components/Timeline/src/types.ts
```diff
@@ -0,0 +1,23 @@
+import BaseProps from '@gemeente-denhaag/baseprops';
+
+export interface TimelineComponent extends BaseProps {
```
ogjtech, line 3: The interface name and file name do not match. Also the interface name is incorrect. I would rename to `IStepTypes` to both indicate an interface and more accurately describe the purpose of the interface.

ogjtech (changes requested):

src/components/Timeline/src/bem-mapping.ts
GewoonMaarten: done

src/components/Timeline/src/StepContent.tsx
GewoonMaarten: I was unable to do this. If I removed this MUI complained `Material-UI: <StepContent /> is only designed for use with the vertical stepper.` Even when I passed all the properties described in the documentation https://material-ui.com/api/step-content/

src/components/Timeline/src/Step.tsx
GewoonMaarten: same story as https://github.com/Gemeente-DenHaag/denhaag-component-library/pull/221#discussion_r650394921

src/components/Timeline/src/StepIcon.tsx
GewoonMaarten: done

src/components/Timeline/src/bem-mapping.ts
GewoonMaarten: They are from Material,  I cannot change them unless I rewrite the component myself

src/components/Timeline/src/styles/step-label.css
GewoonMaarten: These are not overwrites from material, but for the standard browser styling.

src/components/Timeline/src/types.ts
GewoonMaarten: We have never prefixed our interfaces with `I`

src/components/Timeline/src/Timeline.tsx
GewoonMaarten: We have never prefixed our interfaces with `I`.

src/components/Timeline/src/styles/step-content.css
GewoonMaarten: moved

src/components/Timeline/src/Timeline.tsx
GewoonMaarten: done

src/components/Timeline/src/stories/Timeline.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+      </Button>
+    </React.Fragment>
+  );
+};
+
+/**
+ * Implementation of Timeline
+ * "activeStep" property set to enforce step visibility.
+ */
+export const Default = Template.bind({});
```
DanielvanVliet, line 93:
```suggestion
export const Default: Story<TimelineProps> = Template.bind({});
```

DanielvanVliet (changes requested):
Looks good otherwise!

DanielvanVliet (commented):
Does the show code display usable code? Otherwise can you add a custom code string?

GewoonMaarten:
> Does the show code display usable code? Otherwise can you add a custom code string?

@DanielvanVliet Yes, I've made the entire story the code example to highlight its usage.

src/styles/Components/src/tokens-timeline.css
```diff
@@ -0,0 +1,29 @@
...
+  --denhaag-timeline-step-icon-color: var(--denhaag-grey-3);
+  --denhaag-timeline-step-icon-completed-color: var(--denhaag-green-3);
+  --denhaag-timeline-step-icon-margin: auto;
+  --denhaag-timeline-step-icon-padding: 12px;
+  --denhaag-timeline-step-icon-size: 20px;
+  --denhaag-timeline-step-icon-text-active-color: var(--denhaag-white);
+  --denhaag-timeline-step-icon-text-font-size: var(--denhaag-font-size-s);
+
+  --denhaag-timeline-step-label-active-color: var(--denhaag-blue-3);
+  --denhaag-timeline-step-label-active-font-weight: 700;
```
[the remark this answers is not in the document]
GewoonMaarten: This is supposed to be bold. In Figma it states that is should be 700: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=3497%3A16151

src/components/Timeline/src/stories/Timeline.stories.tsx
GewoonMaarten: Added

src/styles/Components/src/tokens-timeline.css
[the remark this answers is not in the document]
GewoonMaarten: See https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn?node-id=3785:35#87864269

src/components/Timeline/package.json
```diff
@@ -18,15 +18,24 @@
...
-        "build": "tsc -b",
+        "build": "tsc -b && copyfiles -u 1 \"src/**/*.css\" dist",
         "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
     },
     "dependencies": {
         "@gemeente-denhaag/baseprops": "^0.1.7",
-        "@material-ui/core": "^4.11.0"
+        "@gemeente-denhaag/design-tokens-common": "^0.1.1",
+        "@gemeente-denhaag/design-tokens-components": "^0.1.1",
+        "@gemeente-denhaag/design-tokens-proprietary": "^0.1.1",
```
DanielvanVliet, line 28: These should not be added here, but installed separately by user of the library.
```suggestion
```

DanielvanVliet (approved):
Otherwise you have my vinkje.

DanielvanVliet (dismissed):

src/components/Timeline/src/StepContent.tsx
mjcarsjens: Alright, nice work on checking it out anyway. Let's just leave it for now and add a issue on the backlog to change this when the next release of MUI is out.

mjcarsjens: That still leaves my comment wrt the className, my apologies, gets a bit confusing when multiple findings are documented in one thread.

src/components/Timeline/src/StepIcon.tsx
mjcarsjens: You're saying done, and yet i'm still seeing two '__' in the classname 😋

src/components/Timeline/src/bem-mapping.ts
mjcarsjens: Same here, doesn't actually seem to be resolved?

mjcarsjens (commented):
Making some nice progress, left some comments in previously opened threads and unresolved two issues as they don't actually seem to be fixed yet

src/components/Timeline/src/StepIcon.tsx
GewoonMaarten: Sorry, I forgot to push my changes

src/components/Timeline/src/Step.tsx
ogjtech: When doing this, do not forget to also add the properties specified in `BaseProps` in the component

src/components/Timeline/src/bem-mapping.ts
GewoonMaarten: Yes, I forgot to push the changes on my laptop.

src/components/Timeline/src/StepLabel.tsx
```diff
@@ -0,0 +1,63 @@
...
+import clsx from 'clsx';
+
+import StepIcon, { StepIconProps } from './StepIcon';
+import { StepComponentProps } from './StepComponentProps';
+
+export interface StepLabelProps extends StepComponentProps {
+  /**
+   * The label displayed in the step icon.
+   */
+  icon?: number;
```
ogjtech, line 11: Why is this property of type `number`? Feel like this should be at least `string | number`...

src/components/Timeline/src/Timeline.tsx
ogjtech: Renaming them is not that difficult, most IDEs have easy rename functions and often do automatic refactoring as well. I know in WebStorm one can press Shift+F6 to rename a c;lass or interface and it automatically refactors almost everything from imports to inherits to package.jsons

src/components/Timeline/src/bem-mapping.ts
ogjtech: The `colorAction`, -`Secondary`, -`Primary` and -`Disable` classes are from Material, yes. But the BEM classes make absolutely zero sense the way they are defined now. It is unclear whether these classes add styling to the `Timeline` component or to the `StepIcon` component. The `stepIconClasses` const itself is already defined inside the `Timeline` component, so specifying this in the BEM-class is redundant and kills reusability. Just `step-icon--disabled` and the like would suffice here in my opinion. @Robbert @mjcarsjens what are your opinions?

ogjtech (changes requested):

src/components/Timeline/src/Timeline.tsx
GewoonMaarten: I know, but no other interface we have is prefixed with `I`

src/components/Timeline/src/StepLabel.tsx
GewoonMaarten: I've updated the type.

src/components/Timeline/src/Step.tsx
GewoonMaarten: I've added `id` to every component

src/components/Timeline/src/StepLabel.tsx
GewoonMaarten: I couldn't get `defaultProps` to work, because it kept overwriting my value with the `defaultProps` value.

DanielvanVliet (approved):

ogjtech (approved):

PR #229, opened 2021-06-05
Add touch event to onclick

bddjong:

### DIFF
--- a/src/components/Button/src/index.tsx
+++ b/src/components/Button/src/index.tsx
@@ -10,7 +10,7 @@ export interface ButtonProps extends BaseProps {
   /**
    * Simple click handler
    */
-  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement> | React.TouchEvent<HTMLButtonElement>) => void;
 
   /**
    * Size of the component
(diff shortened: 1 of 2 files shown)


DanielvanVliet:
Check with storybook whether a mobile touch fires this event.

src/components/Button/src/index.tsx
```diff
@@ -10,7 +10,7 @@ export interface ButtonProps extends BaseProps {
   /**
    * Simple click handler
    */
-  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement> | React.TouchEvent<HTMLButtonElement>) => void;
```
ogjtech, line 13: Instead of using an OR-operand here for every possible event, to include as many events as possible, I would use React's `SyntheticEvent` here. This is a wrapper around pretty much any and all native browser events (which you can obviously restrict by specifying it as a param for the `onClick` property.

ogjtech (changes requested):

src/components/Button/src/index.tsx
GewoonMaarten: `MouseEvent` and `TouchEvent` are more specific than `SyntheticEvent` and include more specific fields which the user can use.

ogjtech (dismissed):

GewoonMaarten (approved):

DanielvanVliet (approved):

joostvanviegen (approved):

PR #248, opened 2021-06-12
Typography design

### LINKED ISSUE #106: Design: Typography
The Typography component needs to be styled according to the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=3497%3A15951).
As with other components, design tokens must be added to achieve this.

We're creating a custom react component instead of wrapping the MUI Typography. Examples from Utrecht
- https://github.com/nl-design-system/utrecht/pull/43
- https://github.com/nl-design-system/utrecht/pull/49

## Acceptance Criteria

**Variants implemented**
- [x] headings
- [x] paragraphs

**Font weights** 
- [x] Default
- [x] Bold
- [x] Font weights as design token

**Requirements**
- [x] Font family can be adjusted (in our designs TheSans and TheMix are used)

DanielvanVliet:
Added heading components and paragraph components according to the figma design.

- Card case was changed to use material-core typography since it makes use of it in its API, and is currently under development #134.
- For headers the sans-serif is used by default, while it should become TheMix. Should the TheMix font be added to the Proprietary component?
- Makes use of the CSS from Utrecht, see the [discussion on Slack](https://codefornl.slack.com/archives/C022AGFTF7Z/p1623828465003700)

Note that most of the design tokens are not properly chosen yet, as Typography is still in progress on Figma.

**Closing issues**

closes #106


src/components/Typography/src/Heading1.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import BaseProps from '@gemeente-denhaag/baseprops';
+
+export type Heading1Props = BaseProps;
+
+export const Heading1: React.FC<Heading1Props> = (props: Heading1Props) => {
```
ogjtech, line 6: Not sure we should use numbers in component names. Either write the numbers as words (`HeadingOne`, `HeadingTwo`, et cetera) or perhaps use a different alias? Maybe something like the Greek alphabet (`HeadingAlpha`, `HeadingBeta`)?

ogjtech:
@DanielvanVliet I noticed you removed the Typography component altogether. Personally, for maintainability sake, I would keep the Typography component, maybe abstract it a little to keep it minimal, and have your variants implement or extend the class. That way you omit duplicate code and can maintain general Typography elements or functions in one place.

src/components/Card/src/stories/Paper.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Meta, Story } from '@storybook/react';
-import Typography from '@gemeente-denhaag/typography';
+import { Typography } from '@material-ui/core';
```
ogjtech, line 3: I would not use the Material UI Typography component, we want our component library to use our own as much as possible.

src/components/Typography/src/typography.module.css
```diff
@@ -0,0 +1,53 @@
...
+  letter-spacing: var(--denhaag-heading-5-letter-spacing);
+}
+
+.denhaag-paragraph {
+  font-size: var(--denhaag-paragraph-font-size);
+  font-family: var(--denhaag-paragraph-font-family);
+  font-weight: var(--denhaag-paragraph-font-weight);
+  font-style: var(--denhaag-paragraph-font-style);
+  line-height: var(--denhaag-paragraph-line-height);
+  letter-spacing: var(--denhaag-paragraph-letter-spacing);
```
Robbert, line 52: Do you really need to specify a custom `letter-spacing`? And if Figma says so: does it really look better that way in the browser?

src/components/Typography/src/typography.module.css
```diff
@@ -0,0 +1,53 @@
...
+}
```
Robbert, line 53: Since we work together towards an implementation of every component that works for most organisations, it would have been very interesting if you would have started by taking the CSS implementation from Utrecht for their Heading and Paragraph component, and worked from there to see if anything was missing.

Perhaps it is not too late to give that a try?

https://github.com/nl-design-system/utrecht/blob/main/components/heading/index.css

https://github.com/nl-design-system/utrecht/blob/main/components/paragraph/index.css

src/components/Typography/src/Heading1.tsx
Robbert: I have never heard of someone who was confused by this, or anyone who upgraded their `<h1>` to `<h6>` to have the latest features.

bddjong: Agreed with @Robbert. This Using Heading[1-6] perfectly reflects the actual HTML element. And makes any design tokens names easy to understand.

src/components/Typography/src/typography.module.css
DanielvanVliet: I have removed it for now, since I agree this likely won't be used. If the Typography Figma is updated we could add it again.

src/components/Card/src/stories/Paper.stories.tsx
DanielvanVliet: The Typography component as used in Card/Paper does not exist anymore, but it still relies on it.
Once we design the paper component this can be removed.

DanielvanVliet:
Waiting for further Figma information about paragraphs. Do we want to expose all 4 paragraphs shown in [Figma](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=3785%3A680) as separate components? Or as react variants on the props of Paragraph

Robbert:
@DanielvanVliet I have a suggestion to only allow 3rd-party prefixes in the design tokens directory, and temporarily also for the Typography component, instead of everywhere in the codebase:

https://github.com/Gemeente-DenHaag/denhaag-component-library/commit/735976d827495b4a40d75a2b4eb6300b518199f7

src/components/Typography/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "composite": true
   },
   "include": ["src/**/*"],
-  "exclude": ["dist", "src/*.stories.tsx"],
+  "exclude": ["dist", "src/*.stories.tsx", "src/**/*.stories.tsx"],
```
bddjong, line 10: `src/**/*.stories.tsx` should be enough. I believe you can remove the exclude in the middle.

bddjong (approved):

ogjtech (approved):

PR #271, opened 2021-06-26
Use our own icons in Button story

GewoonMaarten:


src/components/Button/package.json
```diff
@@ -28,7 +28,7 @@
     "@material-ui/core": "^4.11.0"
   },
   "devDependencies": {
-    "@material-ui/icons": "^4.11.2"
+    "@gemeente-denhaag/icons": "*"
```
joostvanviegen, line 31: I don't think we want unlimited version ranges

joostvanviegen (changes requested):

src/components/Button/package.json
DanielvanVliet: It is a devDependency because it is only used in storybook, where you always want the latest value so it should be fine.

joostvanviegen (approved):

DanielvanVliet (approved):

PR #272, opened 2021-06-26
Implemented design for Menu buttons

### LINKED ISSUE #70: Design: Menu buttons
The Menu Button component needs to be styled according to the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1923%3A4663).

As a base for this we will use the Material-UI `Link` and `PopOver` components instead of the Material-UI `Menu` component.

This will result in two components, one `Menu` component in which multiple `Menu Button` components can exist.

## Acceptance criteria
**Variants implemented**
- [x] With options
- [x] Without options

**States implemented**
- [x] Normal
- [x] Hover
- [x] Focus
- [x] Active

**Animations**
Animations are out of scope at this point.

**Dropdown**
As the dropdown design is not yet complete the dropdown styling will be performed in another issue once the design is finished.

**Requirements**
- [x] Without options variant is represented by a HTML `<a>` component
- [x] The dropdown version is represented by a HTML `<button>` component
- [x] [Alignment](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1923%3A4663)  requirements are met
- [x] Storybook shows HTML code snippet (with classes and accessibility suggestions) (Depends on #267 )

joostvanviegen:
This pull requests adds the Menu Button and Menu component (which is a container for multiple MenuButtons [see figma](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1923%3A4663).

**Closing issues**
This pull request closes #70


src/components/MenuButton/package.json
```diff
@@ -0,0 +1,33 @@
+{
+    "name": "@gemeente-denhaag/menubutton",
+    "version": "0.1.14",
```
bddjong, line 3: Why are we starting on version 0.1.14 instead of 0.1.0

src/components/MenuButton/src/bem-mapping.jsx
```diff
@@ -0,0 +1,7 @@
+export const menuButtonClasses = {
+  root: 'denhaag-menubutton',
```
bddjong, line 2: `denhaag-menu-button` is nicer imo. We've used hyphens where the component itself would have PascalCasing in most places

EDIT: Why is there a bem-mapping file if there's no MUI classes we need to overwrite?

src/components/MenuButton/src/index.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { StylesProvider } from '@material-ui/core/styles';
```
bddjong, line 2: Import the StylesProvider from `@material-ui/core` (PS we don't currently need the styles provider since there's no MUI component being wrapped.)

src/styles/Components/src/tokens-menubutton.css
```diff
@@ -0,0 +1,20 @@
+:root {
+    --denhaag-menubutton-active-color: var(--denhaag-menubutton-hover-color);
```
bddjong, line 2: Same here, I wouldn't mind calling it `--denhaag-menu-button..`

src/styles/Components/src/tokens-menubutton.css
```diff
@@ -0,0 +1,20 @@
...
+    --denhaag-menubutton-active-font-weight: 700;
+
+    --denhaag-menubutton-chevron-active-color: var(--denhaag-menubutton-hover-color);
+    --denhaag-menubutton-chevron-font-size: 12px;
+    --denhaag-menubutton-chevron-hover-color: var(--denhaag-menubutton-chevron--active-color);
```
bddjong, line 7: Something messy going on here. The token your referencing is called `..chevron--active-color` but it doesn't exist. (`chevron-active-color` should be correct

src/styles/Components/src/tokens-menubutton.css
```diff
@@ -0,0 +1,20 @@
...
+    --denhaag-menubutton-color: var(--denhaag-grey-4);
+
+    --denhaag-menubutton-font-family: var(--denhaag-font-family);
+    --denhaag-menubutton-font-size: var(--denhaag-fontsize-medium);
+    --denhaag-menubutton-font-weight: 500;
+
+    --denhaag-menubutton-hover-color: var(--denhaag-blue-3);
+    --denhaag-menubutton-line-height: 21px;
+    --denhaag-menubutton-margin-inline-end: 40px;
+}
```
bddjong, line 20: The ordering of the tokens feels a little awkward. Some token values refer to a token that's defined after itself rather than before itself.

bddjong (changes requested):

src/styles/Components/src/tokens-menubutton.css
joostvanviegen: Fixed

src/styles/Components/src/tokens-menubutton.css
joostvanviegen: I have changed the ordering, thanks for the feedback

src/styles/Components/src/tokens-menubutton.css
```diff
@@ -0,0 +1,49 @@
...
+  * @tokens MenuButton
+  * @presenter FontFamily
+  */
+  --denhaag-menu-button-font-family: var(--denhaag-font-family);
+
+  /**
+  * @tokens MenuButton
+  * @presenter FontSize
+  */
+  --denhaag-menu-button-font-size: var(--denhaag-fontsize-medium);
```
bddjong, line 20: This token does not exist

DanielvanVliet (dismissed):
You forgot to create a Menu component:

Issue #70:
> This will result in two components, one Menu component in which multiple Menu Button 
components can exist.

I believe all it should do is ensure spacing and be responsive.

src/components/Menu/src/menu.css
```diff
@@ -0,0 +1,3 @@
+.denhaag-menu :not(:last-child) {
+      margin-inline-end: var(--denhaag-menu-button-margin-inline-end);
+  }
```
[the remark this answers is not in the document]
bddjong: And I still strongly prefer using `column-gap` since it's much more explicit about what we're doing, instead of adding a margin through a :not selector.

src/components/Menu/src/MenuButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </span>
+      </button>
+    );
+  } else {
+    component = (
+      <a id={props.id} href={props.href} className="denhaag-menu-button">
+        {props.children}
+      </a>
+    );
+  }
```
bddjong, line 42: I think it'll be easier to discern the code for each variant (since they're practically different components) if we do something like this (just a quick type-out, the order is prob not optimal)
```jsx
export const MenuButton: React.FC<MenuButtonProps> = (props: MenuButtonProps) => {
  return props.variant === 'normal' ? menuButton(props) : expandableMenuButton(props);
};

const menuButton = (props: MenuButtonProps) => (
  <a id={props.id} href={props.href} className="denhaag-menu-button">
    {props.children}
  </a>
);

const expandableMenuButton = (props: MenuButtonProps) => (
  <button id={props.id} onClick={props.onclick} className="denhaag-menu-button">
    {props.children}
    <span className="denhaag-menu-button__chevron">
      <ChevronDownIcon aria-label="ChevronDownIcon" />
    </span>
  </button>
);

export default MenuButton;
```

joostvanviegen: Shouldn't we have done this with Card/Case too? I don't think it's a bad idea, but I do think we should do these things consistently

bddjong (approved):

GewoonMaarten (approved):

PR #298, opened 2021-07-06
Feature/#55 checkbox design

### LINKED ISSUE #55: Design: Checkbox
The Checkbox component needs to be styled according to the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1450%3A4570).

## Acceptance criteria
**Variants implemented**
- [x] Active
- [x] Inactive

**States implemented**
- [x] Normal
- [x] Hover
- [x] Disabled
- [x] Error
- [x] Focus

**CheckboxGroup (FormGroup Component)**
- [x] Each checkbox group can have a header label
- [x] Each checkbox group can have a help text
- [ ] Each checkbox group can have an error text which should appear in conjunction with all checkboxes in the group having the Error state (which replaces the help text)

**Requirements**
- [x] Checkmark provided as SvgIcon, not default MaterialUI checkmark
- [x] [Alignment](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1450%3A4570)  requirements are met
- [x] Storybook shows HTML code snippet (with classes and accessibility suggestions) (Depends on #267 )

DanielvanVliet:
**Closing issues**

closes #55


src/components/Checkbox/src/Checkbox.stories.tsx
```diff
@@ -29,41 +29,6 @@ const defaultCode = `
 </FormControlLabel>
```
joostvanviegen, line 29: Shouldn't this be an mdx file (because of documentation purposes)?

DanielvanVliet: Now it is 😀

src/components/Checkbox/src/stories/Checkbox.stories.mdx
```diff
@@ -0,0 +1,399 @@
...
+    parameters={{
+      docs: {
+        source: {
+          code: `
+<FormControlLabel
+  label="Checkbox"
+  control={<Checkbox onChange={() => {}}></Checkbox>}
+></FormControlLabel>
+// This 👆 transpiles into this 👇 in the DOM
+<label class="denhaag-form-control-label">
```
joostvanviegen, line 68: We could probably remove the second part since we also have a pure html example further down. What are your thoughts on this @bddjong>

src/components/FormControlLabel/.stylelintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../../.stylelintrc.json",
+  "rules": {
+    "custom-property-pattern": "(denhaag|utrecht)-[a-z0-9-]+",
```
joostvanviegen, line 4: Wouldn't it be better to just allow this for more components? I don't think this will be the only component which will use utrecht tokens.

src/components/FormControlLabel/src/index.tsx
```diff
@@ -43,7 +44,11 @@ export interface FormControlLabelProps extends BaseClassesProps {
  * FormControlLabels allow the developer to add a label to some components
  */
 export const FormControlLabel: React.FC<FormControlLabelProps> = (props: FormControlLabelProps) => {
-  return <MaterialFormControlLabel {...props} />;
+  return (
+    <StylesProvider injectFirst>
```
joostvanviegen, line 48: I thought this was no longer needed?

src/components/Checkbox/src/stories/Checkbox.stories.mdx
bddjong: Agreed

src/components/FormControlLabel/src/formcontrollabel.css
```diff
@@ -0,0 +1,21 @@
+.denhaag-form-control-label {
+  margin-inline-start: var(--denhaag-form-control-label-margin-inline-start);
+  margin-inline-end: var(--denhaag-form-control-label-margin-inline-end);
+  margin-block-start: var(--denhaag-form-control-label-margin-block-start);
+  margin-block-end: var(--denhaag-form-control-label-margin-block-end);
+}
+
+.denhaag-form-control-label-label {
```
bddjong, line 8:
```suggestion
.denhaag-form-control-label__label {
```

bddjong (changes requested):
One of the HTML variants is disabled, but it's label is not properly styled as disabled. Also, the checkbox story opens the no label variant as default, instead of the default story. Think we can fix that?

bddjong (approved):

PR #299, opened 2021-07-06
icon-button design tokens

DanielvanVliet:
Adds design tokens to IconButton for increased re-usability. The design tokens are based on figma, but icon-button is not an "official" component.

List was adjusted to use the new design tokens, and switched to outline instead of focus border.

Updated IconButton API since some properties were replaced by tokens.


src/styles/Components/src/tokens-icon-button.css
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+  * @tokens IconButton
+  * @presenter Spacing
+  */
+  --denhaag-icon-button-focus-outline-color: var(--denhaag-ocher-4);
+  --denhaag-icon-button-focus-outline-width: var(--denhaag-focus-border-width);
+  --denhaag-icon-button-focus-outline-style: var(--denhaag-focus-border-style);
+
+  --denhaag-icon-button-focus-outline-offset: 0px;
```
bddjong, line 27: Some of the presenters in the tokens file are incorrect. Also, we can use the new and improved inline presenter syntax! Example:

```suggestion
  /**
  * @tokens IconButton
  */
  --denhaag-icon-button-focus-outline-color: var(--denhaag-ocher-4); /* @presenter Color */
  --denhaag-icon-button-focus-outline-width: var(--denhaag-focus-border-width); /* @presenter Spacing */
  --denhaag-icon-button-focus-outline-style: var(--denhaag-focus-border-style); /* @presenter Border */

  --denhaag-icon-button-focus-outline-offset: 0px; /* @presenter Spacing */
```

bddjong (changes requested):

bddjong (approved):

PR #308, opened 2021-07-09
Fix/incorrect imports

### LINKED ISSUE #303: Incorrect imports
Imports from Material UI are not optimized for the rollup build in a lot of components. For example
```
import MaterialStepper from '@material-ui/core/Stepper';
```
should be
```
import { Stepper as MaterialStepper } from '@material-ui/core';
```

This greatly reduces the bundle size, and makes sure that the external dependency is added as a dependency on the end-developers workspace, instead of inlining the required code. The import needs to be from `@material-ui/core` because our `rollup.config.js` marks that module as external dependency.

DanielvanVliet:
Some imports could not be changed to @material-ui/core and thus I
decided to add those to rollup external dependencies.
This way they should not be included ever in the build, even if in the
future more items are imported from those packages.

**Closing issues**

closes #303

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -8,7 +8,15 @@ import postcss from 'rollup-plugin-postcss';
 
 import tsconfig from './tsconfig.json';
 
-const externalDependencies = ['@material-ui/core', '@material-ui/lab', '@material-ui/pickers', 'react', 'react-dom'];
+const externalDependencies = [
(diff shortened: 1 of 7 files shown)


GewoonMaarten (approved):

bddjong (approved):
Good to have that out of the way :D, our internal depencies are working correctly?

PR #315, opened 2021-07-12
Storybook theme switcher and Style Dictionary theme output

Robbert:
Replace "light" and "dark" toolbar button with full-fledged theme switcher addon.

https://github.com/tonai/storybook-addon-themes

To make this work, the design tokens should always not be available on `:root` but be provided via a theme class name instead.

I have added an additional file to the Style Dictionary output, `theme/index.css`, that provides all the design tokens for the `.denhaag-theme` class name.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -4,7 +4,17 @@ module.exports = {
     '../src/components/*/src/**/*.stories.@(js|jsx|ts|tsx)',
     '../src/components/*/src/**/*.stories.mdx',
   ],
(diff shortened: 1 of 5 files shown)


Robbert:
If you think you wouldn't even need the `:root{}` version, I can update this PR to configure the class name on the existing `dist/index.css` output, instead of adding a new file.

bddjong (dismissed):
It's hard to say if there's a need for the :root default export if the library isn't being used yet. I do think it's preferable not having to set a class on the document to get the tokens as a package user

joostvanviegen (dismissed):
I agree with @bddjong on this, I'm not quite sure if and when a `:root` default export is needed.

Issue #154, opened 2021-05-01
Testing components Button and Card/Case


DanielvanVliet:
In the issue [#107 Establishing testing strategy](https://github.com/Gemeente-DenHaag/denhaag-component-library/pull/144), we have proposed a testing strategy which must be implemented on any styled components.

At the time of writing both #54 and #79 are in progress and are not yet ready to be tested.
This issue is about fixing the technical debt of testing for these components, and potentially others that are designed according to NLDS figma.

[Have a peek at the Cypress docs to get familiar with testing components](https://docs.cypress.io/guides/overview/why-cypress)

**Acceptance criteria**
- [x] Button component is tested according our guidelines
- [x] Card/Case component is tested
- [x] For any other component in development designed according to figma, a new issue is created (or is tested within this issue aswell)

PR #368, opened 2021-08-27
Feature/#154 testing completed components

mjcarsjens:
Added automated tests for Button and Card component.
Card component still has an a11y issue with color contrast of the subtitle.

**Closing issues**
#154


bddjong:
Fixed two issues;
- removed an `index.ts` from the card package. (It's supposed to contain only an `index.tsx`),
- removed a file that's no longer being used which imports token css (`src/styles/components/components.module.css`)

Found a bug however. The screenshots cypress creates are sensitive to the system they're run on. Running yarn test on my local machine produces slightly different colors and font weight. This issue is probably the same: https://github.com/jaredpalmer/cypress-image-snapshot/issues/98

Tested it on MacOS with Cypress v8

bddjong:
Also, on one instance, the CI succeeded even though cypress logs show it clearly failed on the card accessibility test. https://github.com/nl-design-system/denhaag/runs/3459361606

src/components/List/specs/List.spec.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+describe(`List tests`, () => {
+  it('has clickable listitems', () => {
+    const props = {
+      onClick: () => {
+        console.log('Clicked');
+      },
+    };
+
+    const spy = cy.spy(props, 'onClick').as('List item click');
```
bddjong, line 13: I noticed we're testing keyboard usage in some components. Might be a good idea to test it here too

src/components/List/specs/List.spec.jsx
```diff
@@ -0,0 +1,58 @@
...
+      {},
+      {
+        selector: '.denhaag-list__item',
+        states: {
+          hover: 'denhaag-list__item--hover',
+          focus: 'denhaag-list__item--focus',
+        },
+      },
+      [{ id: 'list', issue: 402 }],
+    );
```
bddjong, line 56: The List snapshots are suffering from the #402 list issues with a broken layout. Can we change the viewport size to test them properly? https://docs.cypress.io/api/commands/viewport#Defaults. (Or within one specific test suite: https://docs.cypress.io/api/commands/viewport#Reset-viewport-via-Cypress-config)

bddjong (changes requested):

mjcarsjens:
@Robbert fyi: I made some changes during the weekend wrt to the comments you made last friday. 
The new setup uses a git submodule (the `test_images` folder) where all visual regression testing screenshots are stored.

This currently links to the following gitlab repo which is setup with git lfs: https://gitlab.com/mjcarsjens/test_images
The GH action now retrieves all images from the submodule, performs the visual regression tests and adds any missing images (for example for new components) to the submodule repo and updates the submodule ref in this repo. 

Would you mind taking a look at this PR when you have the time to see if you notice anything (about this setup in particular) which should be changed/optimized? 

Thanks in advance!

bddjong:
There is an official cypress github action https://github.com/marketplace/actions/cypress-io

After a quick read I believe it supports caching the generated images/videos in an artifact in case of failure: https://github.com/marketplace/actions/cypress-io#artifacts

src/components/Tab/specs/Tab.spec.jsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import Tab, { TabContext, TabPanel, Tabs } from '../src/index';
+
+describe('Tab tests', () => {
+  // TODO: fix tab tests.
+  /*
+    id and aria-labelledby are set since otherwise they were randomly generated by MUI, which fails the HTML snapshot test.
+    Functionality tests should be added, but due to React.useState/React.setState not working in the tests we need to look for a workaround/fix.
```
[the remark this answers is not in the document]
mjcarsjens: https://github.com/nl-design-system/denhaag/issues/425

src/components/List/specs/List.spec.jsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import List, { ListItem, ListSubheader } from '../src/index';
+import { HouseIcon, TrashIcon } from '../../Icons';
+
+describe(`List tests`, () => {
+  it('has clickable listitems', () => {
```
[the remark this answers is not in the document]
mjcarsjens: I added more tests to properly test both behaviors. Are they extensive enough (for now)?

src/components/List/specs/List.spec.jsx
mjcarsjens: I added some keyboard testing, but quick sidenote: almost all of our keyboard `{enter}` tests currently add absolutely nothing due to https://github.com/nl-design-system/denhaag/issues/428.

src/components/List/specs/List.spec.jsx
mjcarsjens: The issue is in fact not the testing window size (which is wide enough), but the fact that `cy.snapshots` wraps them in the following div: `<div id="wrapper" style="display: inline-block">`. The problem lies within the `display: inline-block`. The `#wrapper` selector is called when creating the screenshots, meaning that always only the actual content of the test is being captured, and not a whole lot of white screen to the right to fill out the testing window.  Due to this the list snapshots look the way they do, but adding `display: block` (or simply removing `inline-block`) will result in all component snapshots filling out the entire testing window (i.e. having a lot of white in them). 

For now I think we should just leave this as is, and create new visual snapshots once this bug is actually fixed.

mjcarsjens: We could technically add some exception where this is not the case, but it feels like a bit of a hassle to add a workaround for a broken component.

mjcarsjens:
@bddjong failed test images (`*.diff.png` and `*.actual.png`) are uploaded in a GH artifact now (example: https://github.com/nl-design-system/denhaag/actions/runs/1262292587)

mjcarsjens:
@bddjong one last quick question: do you think we should also run the visual tests on draft PR's or not?

bddjong:
> @bddjong one last quick question: do you think we should also run the visual tests on draft PR's or not? This is the case right now but just wanna get your thoughts on this.

@mjcarsjens So far I've rarely checked if the actions passed on a draft PR, but since the CI really is a source of truth for the visual regression it might be useful for testing during development... That said, I've no strong preference. Perhaps turn them off and see if we actually miss it? Should be easy to switch back on later, and saves builds on the draft PRs :D 🌲

mjcarsjens:
> 
> 
> > @bddjong one last quick question: do you think we should also run the visual tests on draft PR's or not? This is the case right now but just wanna get your thoughts on this.
> 
> @mjcarsjens So far I've rarely checked if the actions passed on a draft PR, but since the CI really is a source of truth for the visual regression it might be useful for testing during development... That said, I've no strong preference. Perhaps turn them off and see if we actually miss it? Should be easy to switch back on later, and saves builds on the draft PRs :D 🌲

Yea I'm fine with both too, but looking at how many "fix" commits I had one-after-another in a span of minutes just made me wonder, because every one of these commits has to go through these 10 minutes of CI/CD, which seems kinda unnecessary, but when you actually want to test stuff before opening an official PR it definitely will be useful (especially so if you're working on existing components).

bddjong:
Awesome! I only still run into the issue of the date in the card being different from the generated snapshot: 
```
-              <time datetime="Tue Jan 21 2020">1/21/2020</time>
+              <time datetime="Tue Jan 21 2020">21/01/2020</time>
```

mjcarsjens:
> 
> 
> Awesome! I only still run into the issue of the date in the card being different from the generated snapshot:
> 
> ```
> -              <time datetime="Tue Jan 21 2020">1/21/2020</time>
> +              <time datetime="Tue Jan 21 2020">21/01/2020</time>
> ```

Fixed this one by using the `.toISOString()` instead of `.toDateString()` in the `dateTime` prop (which is locale independent).

bddjong (approved):

PR #389, opened 2021-09-10
fix: fixed alignment of icon for link component

### LINKED ISSUE #330: Link component icon not centered
image

Behavior is different between Firefox and Chrome, so be sure to check both browsers.

mjcarsjens:
FIxes the alignment for icons in the Link component.
Bit of a dirty fix imo but due to the padding on the block for the focus border and the icon being higher than the line-height. 

**Closing issues**
closes #330


bddjong (approved):
Looks perfect! Added a small fix for the html-only story, it didn't have the new bem modifier. Also the contrast for the disabled state is insufficient, so I added a comment in figma for that. Not in the scope of this PR though :), ship it!

PR #391, opened 2021-09-10
Card: add new colors and inactive state

bddjong:
This component fixes the accessibility contrast issues with the card, as well as adding the new inactive (gray) variant


mjcarsjens (commented):
Nice! One small comment: is inactive a proper name for this usecase? Maybe we should change this to `archived`

bddjong:
> Nice! One small comment: is inactive a proper name for this usecase? Maybe we should change this to `archived`

Renamed to archived, good suggestion!

mjcarsjens (approved):

PR #418, opened 2021-09-20
Use Utrecht web components for typography

### LINKED ISSUE #352: Use Utrecht NPM packages for Typography
### User story

Our typography components are currently using the Utrecht implementation. However, this was done by creating a 1:1 copy of their CSS, BEM classes and design tokens. We should replace the CSS copy with the CSS from their NPM packages.

bddjong:
WIP, testing with Utrecht typography components in component library

closes #352

### DIFF
--- a/.npmpackagejsonlintrc.json
+++ b/.npmpackagejsonlintrc.json
@@ -14,7 +14,6 @@
     "man-type": "error",
     "name-format": "error",
     "name-type": "error",
-    "prefer-caret-version-dependencies": "error",
     "preferGlobal-type": "error",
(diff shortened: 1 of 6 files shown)


bddjong:
@mjcarsjens I'm curious about your thoughts on this one. Having already used Utrecht design tokens, I was able to swap out the original implementation with the utrecht stencil component without any styling issues. I'm curious about prop forwarding though. Would it be better to directly export the utrecht components so developers have access to its attributes, or is there a neat trick you know of that we can use to expose the utrecht props.

Issue #386, opened 2021-09-08
Link issues


mritense:
### User story

As a developer using the component library, I would like to be able to use the link component

### Describe the solution you'd like

I would like the link component (under Navigation on Storybook) to be included in the component library

### Describe alternatives you've considered

Styling html a element manually

bddjong:
As of version `0.2.3-alpha.29`, the link is still not available when adding the `denhaag-component-library` meta package, even though it's been added to its exports. Should investigate and fix.

**Workaround**
Use `@gemeente-denhaag/link` as extra dependency.

bddjong:
Additionally, the link needs a component prop which allows the user to pass a router library link to handle navigation

PR #440, opened 2021-10-01
Fix/#386 link issues

mjcarsjens:
In this PR the following two things were done:
- Fixed export for link component in `denhaag-component-library` and `navigation` packages.
- Added `RouterLink` component.

A few notes about the `RouterLink` component:
- I experimented with making the `RouterLink` component export a `Link` component from `react-router-dom`. This would however result in issues in the `react-router-dom` implementation; the `Link` component checks whether it is wrapped in a `Router` component **from the same package**. This meant when using our `RouterLink` component in a application with a `Router` component from `react-router-dom` this would throw an error. Because of this we had the following options: export all component from `react-router-dom` ourselves, or adjust our implementation, and I chose to do the latter.
- The `RouterLink` component is supposed to be used as `component` prop in the `react-router-dom` link. When doing this with our default `Link` component this would result in the `react-router-dom` navigation breaking, since when using a custom component there is no `onClick` functionality passed through. Due to this I had to create our own `RouterLink` component with a `onClick` which calls the `navigate()` function which is passed through by the `react-router-dom` `Link` component. This implementation is based on the implentation of the `AnchorLink` component in `react-router-dom` which is used internally when no custom `component` is specified in the `Link`. 

Usage:
`<Link to="/" component={RouterLink} id="123">Home</Link>`

When using this in a TS project there is a small quirk: the `react-router-dom` `Link` component passes on all extra props to the underlying element. However, in its TS definition it only accepts all usable `a` HTMLProps. Due to this you cannot by default use our `icon` and `iconAlign` props of the Link, since these are not default `a` HTMLProps and thus will result in a TS build error, so this will not work:
`<Link to="/" id="123" icon={<ArrowRightIcon />} iconAlign="end">Home</Link>`

To fix this an end-user can do the following:

```
import { RouterLink, LinkProps } from '@gemeente-denhaag/denhaag-component-library';
import { Link as RouterLink, LinkProps as RouterLinkProps } from 'react-router-dom';

interface RouterLinkProps extends Omit<LinkProps, 'href'>, Pick<RouterDomLinkProps, 'to' | 'replace'> {}
const Link: React.FC<RouterLinkProps> = (props: RouterLinkProps) => {
  return <RouterDomLink component={RouterLink} {...props} />;
};
```

By creating a custom `Link` component which accepts both our `LinkProps` and the `to` and `replace` prop from the `LinkProps` from `react-router-dom` you can now do the following in your project: 
`<Link to="/" id="123" icon={<ArrowRightIcon />} iconAlign="end">Home</Link>`

Again, we can not offer this functionality by default out of our component library without also exporting all other `react-router-dom` components a user wants to use, which doesn't seem like a good idea to me, but would like to know your thoughts :)

**Closing issues**
#386

### DIFF
--- a/src/components/BaseProps/src/index.tsx
+++ b/src/components/BaseProps/src/index.tsx
@@ -49,3 +49,18 @@ export interface BaseClassesProps {
    */
   classes?: Record<string, unknown>;
 }
+
+type OverrideProps<M, C extends React.ElementType> = M & Omit<React.ComponentPropsWithRef<C>, keyof M>;
+
+export interface OverridableComponent<M> {
+  <C extends React.ElementType>(
+    props: {
+      /**
+       * The component used for the root node.
+       * Either a string to use a HTML element or a component.
+       */
+      component: C;
+    } & OverrideProps<M, C>,
+  ): JSX.Element;
+  (props: BaseProps & M): JSX.Element;
+}
--- a/src/components/Link/package.json
+++ b/src/components/Link/package.json
@@ -25,9 +25,6 @@
         "@gemeente-denhaag/baseprops": "^0.2.3-alpha.33",
         "@gemeente-denhaag/icons": "^0.2.3-alpha.33"
     },
-    "devDependencies": {
-        "@gemeente-denhaag/typography": "^0.2.3-alpha.33"
-    },
     "peerDependencies": {
         "react": "^17.0.1"
     }
--- a/src/components/Link/src/Link.stories.mdx
+++ b/src/components/Link/src/Link.stories.mdx
@@ -20,6 +20,53 @@ import readme from '../README.md';
 
 <Description markdown={readme} />
 
+## Usage
+This component can be used in two ways:
+
+- As a standard `a`-tag with icon support. For this you need to supply at least an `href`, and an icon can be added in the `icon` prop:
+  `<Link href="/home" icon={<ArrowRightIcon />} iconAlign="end">Home</Link>`
+
+- With another way of routing, for example using `react-router-dom`. For this we have exposed the `component` property. When adding a `component` all props of this component will be usable on our `Link` component and will be passed through.
+  An example with `react-router-dom`:
+
+  ```
+  import { Link } from '@gemeente-denhaag/denhaag-component-library';
+  import { ArrowRightIcon } from '@gemeente-denhaag/icons';
+  import {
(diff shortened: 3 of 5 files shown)


mjcarsjens:
@mritense do you think this solution will be usable?

mritense:
> @mritense do you think this solution will be usable?



> @mritense do you think this solution will be usable?

While looking at the description and the changes I cannot figure out completely what's going on. Let's discuss this during/after call at 2pm

mritense (approved):

mjcarsjens:
@mritense had a bit of a discussion about this solution with Bryan and decided to try make it a bit prettier. It now works without declaring an extra link component in the application. Now it works as follows (we also use the default Link component instead of a RouterLink component now):

```
import { Link } from '@gemeente-denhaag/denhaag-component-library';
import { Link as RouterLink } from 'react-router-dom';

[... some code until you reach the point where you want to use a Link ...]
 
<Link component={RouterLink} to="/" icon="{<ArrowRightIcon />}">Home</Link>
```

As you can see we now give the `Link` component from `react-router-dom` to our `Link` instead of the other way around. When you pass a `component` prop into our `Link` component any props of this component will become available in our component and will be passed on, resulting in no TS errors in this scenario and no need for extra custom code for the end-user (in this case: you).

mritense:
> @mritense had a bit of a discussion about this solution with Bryan and decided to try make it a bit prettier. It now works without declaring an extra link component in the application. Now it works as follows (we also use the default Link component instead of a RouterLink component now):
> 
> ```
> import { Link } from '@gemeente-denhaag/denhaag-component-library';
> import { Link as RouterLink } from 'react-router-dom';
> 
> [... some code until you reach the point where you want to use a Link ...]
>  
> <Link component={RouterLink} to="/" icon={<ArrowRightIcon />}>Home</Link>
> ```
> 
> As you can see we now give the `Link` component from `react-router-dom` to our `Link` instead of the other way around. When you pass a `component` prop into our `Link` component any props of this component will become available in our component and will be passed on, resulting in no TS errors in this scenario and no need for extra custom code for the end-user (in this case: you).

Awesome!

bddjong (approved):

bddjong (approved):
Can we add some documentation about this new property in storybook? An example snippet using `react-router-dom` would be great

bddjong (approved):

Issue #483, opened 2021-11-09
Docs: Getting started


bddjong:
### User story

As a developer using the component library, I want to be able to find a getting started guide so I can use the components without investing a lot of time into project setup.

### Describe the solution you'd like

A "Getting started" guide in storybook. The guide should be easily found (one of the first pages in the storybook documentations).

### Describe alternatives you've considered

A GitHub readme (README.md). This is what we currently have, however, a readme that can only be found on Github makes more sense for contributors than users of the library.

### Additional context

Created due to a question on Slack about usage of the library.

### Acceptance criteria

- [ ] The getting started guide can be easily found on [storybook] (https://nl-design-system.github.io/denhaag/?path=/story/den-haag-introduction--page)
- [ ] Walking through the getting started guide results in a working project with a number of example components from the library
- [ ] The guide explains the usage of design tokens the change the organisation theme

## Definition of done

- [ ] Code exists on master branch / deployed
- [ ] Meets acceptance criteria
- [ ] Complies with [the standard of public code](https://standard.publiccode.net/)

PR #484, opened 2021-11-09
docs: add getting started documentation

bddjong:
Closing issues #483

### DIFF
--- a/.storybook/stories/Introduction.stories.mdx
+++ b/.storybook/stories/Introduction.stories.mdx
@@ -3,20 +3,18 @@ import { Meta } from "@storybook/addon-docs";
 <Meta title="Den Haag/Introduction" />
 
 # Den Haag Design System
+This project is very much WORK IN PROGRESS and all components are released as alpha version. Always define the exact version you want to use, and test for breaking changes before upgrading to a newer alpha release.
+
+Applying design elements from this project is strictly prohibited for organisations that are not part of the Municipality of The Hague. However, we do welcome contributors and users from other organisations. In order to apply your own theme to our components, we support [Design Tokens](?path=/docs/den-haag-design-tokens--page).
+
+This project is part of a community initiative to use NL Design System components for projects that need to adhere to the Den Haag Design System.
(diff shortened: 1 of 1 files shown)


.storybook/stories/Introduction.stories.mdx
```diff
@@ -21,3 +21,41 @@ The design tokens are [available as CSS variables on GitHub](https://github.com/
...
+In order to add the entire component library, install [@gemeente-denhaag/denhaag-component-library](https://www.npmjs.com/package/@gemeente-denhaag/denhaag-component-library) using your favorite package manager.
+
+For example, ```npm install @gemeente-denhaag/denhaag-component-library```.
+
+Alternatively, you can install specific components: ```npm install @gemeente-denhaag/button```.
+
+After installing the dependency, you can start using the Den Haag components as follows:
+```jsx
+import {Button, StylesProvider} from "@gemeente-denhaag/denhaag-component-library";
+// Import your design tokens here as well! Otherwise your components will be unstyled :(
```
mjcarsjens, line 42: Solid smiley usage 👍

.storybook/stories/Introduction.stories.mdx
```diff
@@ -21,3 +21,41 @@ The design tokens are [available as CSS variables on GitHub](https://github.com/
...
+export default App;
+```
+**Important:** Your app, or the section which uses Den Haag components, will need to be wrapped in the `StylesProvider` component. This ensures that styles are loaded correctly.
+
+#### Applying your theme
+
+By default, all components in the library are unstyled.
+If you wish to apply your organisations theme to the components, define the design tokens for the components you need, and import them in your app.
+
+For Den Haag applications, import the [@gemeente-denhaag/design-tokens-components](https://www.npmjs.com/package/@gemeente-denhaag/design-tokens-components) package, and import the css in your application.
```
mjcarsjens, line 61: Maybe also add an example here of how to import these?

bddjong: Added examples for js import as well as regular html head import

Issue #515, opened 2021-11-25
Add css-only package


bddjong:
We need a package which exports only the css of components.

We will be using the [nl-design-system/utrecht](https://github.com/nl-design-system/utrecht) setup for now with one big css package which imports all component css.

Issue #492, opened 2021-11-11
CSP Headers issues


bddjong:
Inline styles can be removed by changing the rollup config to extract css. 

Components using Material UI are fixed when MUI is removed and they're made from scratch.

mjcarsjens:
Adding `extract: true` to the global rollup config does indeed extract the CSS to seperate `.css` files instead of inlining it, however it breaks the visual regression tests for the following components: 

- Alert
- BadgeCounter
- Checkbox
- DotIndicator
- FormControlLabel
- FormGroup

We need to further investigate what goes wrong here and fix these issues.

bddjong:
An alternative solution is to stop shipping CSS with the React Components, and instead expect developers to use `@gemeente-denhaag/components-css` to import styles. 

This will be a nuisance when using a single component package, but maybe we should consider dropping support for single component packages. From what I know, they have so far only been used as a workaround for bugs in the `components-react` package.

YourivHDenHaag:
For now this is fixed with nonces.

PR #521, opened 2021-11-25
#515 css only packages

mjcarsjens:
Added a CSS-only package. This contains 3 files:
- index.css (component css + DH design tokens)
- bem.css (component css only)
- root-theme.css (DH design tokens only)

Also adds `extract: true` to `postcss` plugin in `rollup.config.js` since we need the `.css` files here, but this also solves the first part of #492 .

**Closing issues**
#515


mjcarsjens:
@bddjong Fixed the broken visual regression tests. Some context:

Adding `extract: true` to the `postcss` plugin in rollup seems to break some styling for a few components, resulting in the failed visual regression tests. For now I've removed the `extract: true` from our rollup config, duplicated the rollup config to the css package folder and added `extract: true` to that specific config. Once we've worked out why the styling breaks for these specific components and we've re-added the `extract: true` to the global rollup config we can remove the specific one and point to the global file again.

For now fixing this seems like out of scope for this specific issue though, which is why I went with this approach.

Robbert:
I was thinking of renaming `bem.css` into `index.css`. I think it makes sense, given the design token architecture, the tokens are not bundled by default.

```js
import '@denhaag/component-library';
import '@utrecht/design-tokens';
```

This code above looks like it should work, but would give weird results because there are now two sets of design tokens in the page. The following code could be the "fix", but I think I would rather make the components without design tokens the default instead:

```js
import '@denhaag/component-library/dist/bem';
import '@utrecht/design-tokens';
```

@mjcarsjens What do you think?

mjcarsjens:
> I was thinking of renaming `bem.css` into `index.css`. I think it makes sense, given the design token architecture, the tokens are not bundled by default.
> 
> ```js
> import '@denhaag/component-library';
> import '@utrecht/design-tokens';
> ```
> 
> This code above looks like it should work, but would give weird results because there are now two sets of design tokens in the page. The following code could be the "fix", but I think I would rather make the components without design tokens the default instead:
> 
> ```js
> import '@denhaag/component-library/dist/bem';
> import '@utrecht/design-tokens';
> ```
> 
> @mjcarsjens What do you think?

@Robbert I think that's a fair point, it indeed would make it more intuitive to use when the design tokens are not by default bundled in the `index.css`. Do we want to include the design-tokens at all in this package (as a design-tokens only package already exists) or just leave them seperated? I feel like maybe this package should not include the design-tokens at all and just leave it up to the user to decide which token package to use.

mjcarsjens:
@Robbert I just made a commit of what it would look like with only the components css and no design-tokens in this package.

Robbert:
@mjcarsjens Looks good! Only thing I'm wondering is if having `dist/index.css` is enough, or if you actively want to provide and maintain the `.js` and `.d.ts` files as well. If they are just a side effect from using rollup, you could consider putting `*.js` and `*.ts` in `.npmignore`.

mjcarsjens:
@Robbert you're right, in the previous version it indeed did only build 3 `.css` files but seems like I got a bit over-enthusiastic with the rewriting. Changed the build process so now only a single  `.css` file appears in the `dist` folder. Tested it in another `tsx` file and as long as the `main` is set to the `dist/index.css` file in `package.json` everything works as intended.

mjcarsjens:
@bddjong any thoughts?

packages/component-library-css/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@gemeente-denhaag/component-library-css",
+  "version": "0.2.3-alpha.50",
```
bddjong, line 3: I'm guessing this package isn't starting on alpha version 50 ;o

packages/component-library-css/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "A CSS library for the components from the Municipality of the Hague",
+  "bugs": "https://github.com/nl-design-system/denhaag/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nl-design-system/denhaag.git",
+    "directory": "src/"
```
bddjong, line 9:
```suggestion
    "directory": "packages/component-library-css"
```

packages/component-library-css/rollup.config.js
```diff
@@ -0,0 +1,46 @@
...
+  '@material-ui/core',
+  '@material-ui/lab',
+  '@material-ui/pickers',
+  'react',
+  'react-dom',
+  '@material-ui/core/transitions',
+  '@material-ui/core/OverridableComponent',
+];
+const internalDependencies = _.keys(tsconfig.compilerOptions.paths);
+const dependencies = externalDependencies.concat(internalDependencies);
```
bddjong, line 21: Do we actually need a separate rollup config for this package? In any case, I believe there shouldn't be need for the external code if we're only exporting css from this package.

bddjong (changes requested):

packages/component-library-css/package.json
Robbert: I make the same mistake all the time 🙈

packages/component-library-css/rollup.config.js
mjcarsjens: > @bddjong Fixed the broken visual regression tests. Some context:
> 
> Adding `extract: true` to the `postcss` plugin in rollup seems to break some styling for a few components, resulting in the failed visual regression tests. For now I've removed the `extract: true` from our rollup config, duplicated the rollup config to the css package folder and added `extract: true` to that specific config. Once we've worked out why the styling breaks for these specific components and we've re-added the `extract: true` to the global rollup config we can remove the specific one and point to the global file again.
> 
> For now fixing this seems like out of scope for this specific issue though, which is why I went with this approach.

This comment explains the need for the seperate rollup config, probably could indeed use some changes but as it's not intended to be used long-term not sure whether this is worth our time.

bddjong (approved):

packages/component-library-css/package.json
mjcarsjens: Whoops, is fixed now

bddjong (approved):

PR #523, opened 2021-11-26
#514 deprecate unused meta packages

### LINKED ISSUE #514: Deprecate unused meta packages (input, layout, datadisplay etc)
Deprecate the following meta-packages and remove code from repo:

- [x] datadisplay
- [x] input
- [x] layout
- [x] navigation
- [x] surfaces

mjcarsjens:
Deprecated unused meta-packages, and adjusted denhaag-component-library package accordingly.

**Closing issues**
closes #514


packages/denhaag-component-library/package.json
```diff
@@ -22,13 +22,31 @@
...
     "@gemeente-denhaag/pickers": "^0.2.3-alpha.50",
-    "@gemeente-denhaag/stylesprovider": "^0.1.1-alpha.50",
-    "@gemeente-denhaag/surfaces": "^0.2.3-alpha.50"
+    "@gemeente-denhaag/link": "^0.2.3-alpha.50",
+    "@gemeente-denhaag/menu": "^0.2.3-alpha.50",
+    "@gemeente-denhaag/tab": "^0.2.3-alpha.50",
+    "@gemeente-denhaag/timeline": "^0.2.3-alpha.50",
+    "@gemeente-denhaag/accordion": "^0.2.3-alpha.50",
+    "@gemeente-denhaag/card": "^0.2.3-alpha.50",
+    "@gemeente-denhaag/stylesprovider": "^0.1.1-alpha.50"
```
bddjong, line 49: Can we make sure these are sorted alphabetically? npm-package-json-lint should be warning us about that with the following rule https://npmpackagejsonlint.org/docs/en/rules/dependencies/prefer-alphabetical-dependencies#example-npmpackagejsonlintrc-configuration

packages/denhaag-component-library/src/index.tsx
```diff
@@ -1,6 +1,27 @@
...
+export * from '@gemeente-denhaag/radio';
+export * from '@gemeente-denhaag/select';
+export * from '@gemeente-denhaag/switch';
+export * from '@gemeente-denhaag/textfield';
+export * from '@gemeente-denhaag/menu';
+export * from '@gemeente-denhaag/timeline';
+export * from '@gemeente-denhaag/tab';
+export * from '@gemeente-denhaag/link';
+export * from '@gemeente-denhaag/accordion';
+export * from '@gemeente-denhaag/card';
```
bddjong, line 26: ditto

bddjong (approved):
Small comments about sorting. Also, we may want to rename the `denhaag-component-library` folder to just `component-library` if you prefer a shorter name. Although I think changing the npm package name is probably not such a good idea.

mjcarsjens:
@bddjong changed order around as requested, and added the packagejson lint rule. 

Also changed around the linting setup a bit so folders other than `components` also get linted. 
For stylelint I ran into some issues using `*/**/*.css`, even with correct `.stylelintignore` settings, so I had to resort to specifying the `components`, `proprietary` and `packages` folders here in the command.

mjcarsjens:
> Small comments about sorting. Also, we may want to rename the `denhaag-component-library` folder to just `component-library` if you prefer a shorter name. Although I think changing the npm package name is probably not such a good idea.

Not sure about this one. Agree we should not just go around renaming npm packages, but for consistensies sake I would prefer the folder name to correspond to the package name, even though I agree the name `component-library` would be nicer.

Robbert:
For the example repo I dropped the `library` entirely, and I would just have `components` / `components-react` / `components-css` / `web-components-react`, et cetera: https://github.com/nl-design-system/example/tree/main/packages

Also more in line with how the directory in the repo itself is called. I can totally rename the one from Utrecht as well. Better now than later.

bddjong (approved):

Issue #516, opened 2021-11-25
Write BEM classes component (Button)


bddjong:
Refactor the `Button` component to not use MUI.

**Acceptance criteria**
- [x] No more MUI
- [x] Storybook story exists
- [x] Cypress Test for component
- [x] Passes tests: visual regression (not HTML snapshot)

PR #525, opened 2021-11-26
refactor: remove mui from button

bddjong:
**Closing issues** #516


components/Button/src/index.tsx
```diff
@@ -33,47 +31,34 @@ export interface ButtonProps extends BaseProps {
...
-    case 'secondary-action':
-      muiVariant = 'outlined';
-      break;
-  }
+export const Button: React.FC<ButtonProps> = ({ iconAlign = 'start', ...props }: ButtonProps) => {
+  const rootClassNames = clsx('denhaag-button', {
+    'denhaag-button--secondary-action': props.variant === 'secondary-action',
+    'denhaag-button--large': props.size === 'large',
+    'denhaag-button--start-icon': iconAlign === 'start' && props.icon !== undefined,
+    'denhaag-button--end-icon': iconAlign === 'end',
```
mjcarsjens, line 52: Seems a bit inconsistent to only check if icon is undefined for the start class

components/Button/src/index.tsx
```diff
@@ -33,47 +31,34 @@ export interface ButtonProps extends BaseProps {
...
   /**
-   * Icon placed at the start of the button
+   * Icon to display at the start or the end of the button
    */
-  startIcon?: React.ReactNode;
+  icon?: React.ReactElement<SvgIconProps>;
 
   /**
-   * Icon placed at the end of the button
+   * If an `icon` is specified, should it be aligned on the left or the right?
```
mjcarsjens, line 39: Might want to add that `start` is the default here

components/Button/src/button.css
```diff
@@ -1,129 +1,144 @@
...
+    var(--denhaag-button-border-width)
+  );
+  padding-inline-end: calc(
+    var(--denhaag-button-medium-size-padding-inline, var(--denhaag-button-padding-inline)) -
+    var(--denhaag-button-border-width)
+  );
+  padding-inline-start: calc(
+    var(--denhaag-button-medium-size-padding-inline, var(--denhaag-button-padding-inline)) -
+    var(--denhaag-button-border-width)
+  );
```
mjcarsjens, line 74: Could you explain this one? Seems a bit counter-intuitive to me that when I specify a padding of a certain value the padding set will be that minus the border-width. Not sure this is really user-friendly.

mjcarsjens: Ah I see, we already had that one to ensure no size differences between the secondary action and the primary action button (which has no border), nvm me then

components/Button/src/button.css
```diff
@@ -1,129 +1,144 @@
...
+  line-height: 1.75;
+  padding-block-end: var(--denhaag-button-medium-size-padding-block, var(--denhaag-button-padding-block));
+  padding-block-start: var(--denhaag-button-medium-size-padding-block, var(--denhaag-button-padding-block));
+  padding-inline-end: var(--denhaag-button-medium-size-padding-inline, var(--denhaag-button-padding-inline));
+  padding-inline-start: var(--denhaag-button-medium-size-padding-inline, var(--denhaag-button-padding-inline));
+  position: relative;
 }
 
 .denhaag-button--focus:after,
 .denhaag-button:focus:after {
```
mjcarsjens, line 20: I really don't like the CSS2 `:after` notation, think the readability is much better when using the CSS3 `::after` notation.

mjcarsjens: (Same goed for other usages of the `:after` pseudoselector)

mjcarsjens (commented):
@bddjong Few small comments

PR #530, opened 2021-11-26
ci: add npm-check-updates

Robbert:
Having the `npm-check-updates` plugin can make it easier to install updates suggested by Dependabot, avoid merge conflicts in those PRs, and avoid updating only one package at the time instead of updating a whole bunch of related packages (`@storybook/...-addon` for example).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -245,11 +245,11 @@ it("does not violate any accessibility rules", () => {
 
 Steps to go through:
 
-1. Create a new branch from the latest version of `master`:
-2. `yarn update-patch` to install the latest available patch version of every `dependency` and `devDepenency` of every `package.json` in this npm workspace. E.g.: from 1.0.0 to 1.0.7. Check the logs to see what packages have been affected, and where to perhaps pay speciall attention to when testing for regressions.
+1. Create a new branch from the latest version of `main`:
+2. `yarn update-patch` to install the latest available patch version of every `dependency` and `devDepenency` of every `package.json` in this npm workspace. E.g.: from 1.0.0 to 1.0.7. Check the logs to see what packages have been affected, and where to perhaps pay special attention to when testing for regressions.
 3. Perform "smoke testing", for example using `yarn clean && yarn lint && yarn build`.
 4. `git commit` these updates, so in case of regressions it will be easier to pinpoint what upgrade has caused issues.
-5. `yarn update-minor` to install the latest available minor version of every depependency. E.g.: from 1.0.0 to 1.2.3.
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -240,3 +240,16 @@ it("does not violate any accessibility rules", () => {
   cy.checkA11y("button");
 });
 ```
+
+## Updating dependencies
+
+Steps to go through:
+
+1. Create a new branch from the latest version of `master`:
```
mjcarsjens, line 248: Might as well call it `main` here as that PR is also opened

mjcarsjens (commented):
One small comment, looks good otherwise but haven't had time to test it just yet

CONTRIBUTING.md
```diff
@@ -240,3 +240,16 @@ it("does not violate any accessibility rules", () => {
...
+2. `yarn update-patch` to install the latest available patch version of every `dependency` and `devDepenency` of every `package.json` in this npm workspace. E.g.: from 1.0.0 to 1.0.7. Check the logs to see what packages have been affected, and where to perhaps pay speciall attention to when testing for regressions.
+3. Perform "smoke testing", for example using `yarn clean && yarn lint && yarn build`.
+4. `git commit` these updates, so in case of regressions it will be easier to pinpoint what upgrade has caused issues.
+5. `yarn update-minor` to install the latest available minor version of every depependency. E.g.: from 1.0.0 to 1.2.3.
```
bddjong, line 252: > 5. `yarn update-minor` to install the latest available minor version of every de**pepe**ndency. E.g.: from 1.0.0 to 1.2.3.
image

bddjong (approved):
Works very nice! Could you elaborate on how this should play together with dependabot and why some packages should reject minor updates (such as typescript)?

Robbert:
> @Robbert Works very nice! Could you elaborate on how this should play together with dependabot and why some packages should reject minor updates (such as typescript)?

I guess I left TypeScript in more as an example; for utrecht we have it locked probably because Angular was particular about the version. 

I reckon we should have some documentation per package when to update and what you need to pay attention to.

Robbert:
> @Robbert Works very nice! Could you elaborate on how this should play together with dependabot


Dependabot is still great for the convenience to see the CI results per package in isolation, and spot which upgrades are causing issues.

Ideally the dependabot.yml is generated from the ncu config files too, so dependabot will follow the same patch/minor/major upgrade strategy for specific packages. For utrecht the dependabot.yml is manually maintained and is probably out to date. It does contain code comments about the WHY of fixing package versions.

Robbert:
Also, I always use dependabot PRs to read the changelogs between versions.

PR #540, opened 2021-11-29
ci: update linting with latest from other NL Design System repos

Robbert:
One of the main things I updated is the refactoring of `.eslintrc` into separate files, to avoid plugins being run on file types they don't know. JSON files now can be linted again, the TypeScript plugin no longer checks `.js` files (rendering many `eslint-disable-next-line` comments obsolete), and `mdx` files can now be checked. (The MDX plugin is finnicky with non-mdx files.)

The main things I didn't include yet are the plugins that check import order and the SCSS linting.


Robbert:
@bddjong PR has been updated with the tweaks we decided on. After the fixed #553 has been merged, the `format package.json files with Prettier (2 instead of 4 spaces)` commit has to be done over again, but that should be easily.

PR #552, opened 2021-11-30
Feature/html link

paulAcato:
Adding the HTML preview and rewrite minor CSS variables.


bddjong:
@paulAcato Great start! :D I know its still a draft, so just a few small pointers:

- I think we can call the stories files `bem.stories.mdx` and `react.stories.mdx`. This is more in line with Utrecht. (For the button refactor we recently merged, I'll use the same naming approach.)
- For commit messages we're using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary). Long story short, you can prefix your commit messages with `fix: ...`, `feat: ...`, `ci: ...` and more. Whichever best describes the sort of change. For refactoring I tend to use `refactor: ...` as well.

components/Link/src/link.css
```diff
@@ -24,25 +24,25 @@
...
 .denhaag-link--focus,
 .denhaag-link:focus {
-  color: var(--denhaag-link-focus-color);
+  --denhaag-link-color: var(--denhaag-link-focus-color);
   outline: var(--denhaag-link-focus-outline);
 }
 
 .denhaag-link--hover,
 .denhaag-link:hover {
+  --denhaag-link-color: var(--denhaag-link-hover-color, var(--denhaag-link-color));
```
bddjong, line 33: Nice improvement. One small question: do we need to fallback the variable to itself? I think we can omit the fallback since it would already be set.

mjcarsjens: Furthermore, while I get the the idea here and it would be nice if this worked, CSS variables do not support falling back to its own value. When `--denhaag-link-hover-color` is undefined this would result in `--denhaag-link-color` also becoming undefined, regardless of its previous value.

paulAcato: @bddjong & @mjcarsjens , i have removed the fallback since i noticed it's indeed the same as the previous which is indeed not supported by the browser.

paulAcato:
Closed the Pull request since I first need to fix the merge-conflict with main branch.

PR #562, opened 2021-12-02
Feature/#560 lerna carret dependencies

### LINKED ISSUE #560: Lerna version creates carret dependencies
Lerna version creates carret dependencies at the moment, resulting in invalid package.json files. Whenever you want to make a commit with manual changes to a package.json the linter will now stop you since these carret dependencies are not allowed.

mjcarsjens:
Added `exact` flag to both `lerna version` and `lerna add`. Furthermore changed the commit message for `lerna version` to be the same as for other NLDS repos and now only allow `lerna version` from the `main` branch.

**Closing issues**
closes #560


bddjong:
@mjcarsjens We should probably fix the incorrectly bumped internal dependencies in this PR (caret that was placed back by previous lerna version)

mjcarsjens:
> @mjcarsjens We should probably fix the incorrectly bumped internal dependencies in this PR (caret that was placed back by previous lerna version)

image

PR #568, opened 2021-12-02
Feature/html link

bddjong:
**Closing issues**


paulAcato:
@mjcarsjens , just rebased and merge @bddjong the main branch into the feature branch ;-)

components/Link/src/link.css
```diff
@@ -24,25 +24,22 @@
...
-}
-
-.denhaag-link--disabled {
-  color: var(--denhaag-link-disabled-color);
-  cursor: none;
-  pointer-events: none;
 }
 
+.denhaag-link--disabled,
 .denhaag-link--disabled--hover,
```
mjcarsjens, line 38: So I know this one was already in here, but looking at this from a BEM perspective having double modifier identifiers (`--`) is not really up to spec. Furthermore, these `--hover` and `--focus` classes are used for visual regression testing. However, as `.denhaag-link--disabled` is already below `.denhaag-link--hover` in the CSS having this extra one is not necessary, as any link with both `denhaag-link--disabled` and `denhaag-link--hover` classes will already have the disabled color.

I would suggest just removing this line as it is not necessary here and just going with 
```
.denhaag-link--disabled,
.denhaag-link--disabled:hover {
```

(The `':hover` is still necessary since that pseudoselector makes the rule more specific).

components/Link/src/stories/bem.stories.mdx
```diff
@@ -0,0 +1,214 @@
...
+            </Route>
+            <Route path="/">
+              <h1>Home</h1>
+            </Route>
+          </Switch>
+      </div>
+    </Router>
+  );
+}
+```
```
mjcarsjens, line 66: This documentation block is very React specific, should probably just be removed from the HTML story.

mjcarsjens (commented):
2 small comments, but overall nice setup with the two seperate stories!

components/Link/src/link.css
mjcarsjens: Small other comment, we might want to consider actually changing this to 

```
.denhaag-link--disabled,
.denhaag-link--disabled.denhaag-link--hover,
.denhaag-link--disabled:hover {
```

I honestly think all uses of the  `--hover` and `--focus` should be replaced in this file, so for example the rule `.denhaag-link--hover` would become `.denhaag-link.denhaag-link--hover`. This way we ensure that the specificity of the default `:hover` rule is the same as for the visual regression testing `--hover` rule. 

In this current implementation there could be scenario's where styling in the testing `--hover` rule gets overridden while not the case for the actual `:hover` rule. This would make the results of the visual regression tests inconsistent with what actually is shown to the end-user. I know we do this in (at least a few) other components, might want to create a seperate issue for this to check the other components if this is a wider issue or only in this component.

@bddjong thougts?

paulAcato: @mjcarsjens , I agree with you here, I didn't change the CSS selectors since I was told this was used for the regression tests and I don't know if it will break anything where it's currently used.

components/Link/src/stories/bem.stories.mdx
paulAcato: Just removed it, think we looked over this react code in the bem stories when we did a git rebase

mjcarsjens: Nice!

components/Link/src/link.css
mjcarsjens: Created a seperate issue to check this in all components and change it where necessary: #580

mjcarsjens: For now this change is out of scope for this PR, so this one can be merged!

mjcarsjens (approved):

components/Link/src/link.css
bddjong: @mjcarsjens Agreed, and good idea to separate this into a different issue for more components.

robbieatacato:
# Documentation and design:
 https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1485%3A1156

PR #570, opened 2021-12-02
fix: add missing dependency

### LINKED ISSUE #569: fix @material-ui/lab dependency missing
Using the component library package will now result in a missing dependency error being thrown due to `@material-ui/lab` not being installed. The problem here lies with the fact that our `Tab` component uses `@material-ui/lab` but does not supply this dependency in its `package.json`. Until now this went unnoticed due to the fact that it was in fact declared in the `navigation` meta-package `package.json`.  However, now the meta-packages have been removed, this bug comes to light.

mjcarsjens:
**Closing issues**
closes #569

### DIFF
--- a/components/Tab/package.json
+++ b/components/Tab/package.json
@@ -23,7 +23,8 @@
   },
   "dependencies": {
     "@gemeente-denhaag/baseprops": "0.2.3-alpha.62",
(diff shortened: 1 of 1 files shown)


bddjong (approved):
Good spot

PR #584, opened 2021-12-03
fix(button): use react button props

bddjong:
In the button refactoring, prop forwarding was lost when MUI was removed from the component. This PR restores its function.

Additionally, it uses the default React interface for HTML Button attributes. This means we don't have to manually support all the default HTML attributes such as `id`, `disabled`, `type` and more!


mjcarsjens:
@joostvanviegen to the rescue

mjcarsjens:
guess not :(

components/Button/src/index.tsx
```diff
@@ -44,17 +33,22 @@ export interface ButtonProps extends BaseProps {
...
+      'denhaag-button--end-icon': iconAlign === 'end',
+    },
+    props.className,
+  );
+
   const iconWrapped = props.icon !== undefined ? <span className={'denhaag-button__icon'}>{props.icon}</span> : '';
 
   return (
-    <button className={rootClassNames} onClick={props.onClick} disabled={props.disabled} type={props.type}>
+    <button {...props} type={type} className={className} onClick={props.onClick}>
```
mjcarsjens, line 51: Is it necessary to have onClick written out here (as it is also covered by the spreaded `props`)?

mjcarsjens (approved):
One comment but it works

components/Button/src/index.tsx
mjcarsjens: Also feel a bit weirdly about the `className` being set by the `props` and then overwritten by specifying it to the `className` const, but guess it works

bddjong: Fixed

bddjong:
`type` is now defaulted to `button`. However, all snapshots tests with a button are failing because of this change.

PR #586, opened 2021-12-06
Feature/breadcrumbs

### LINKED ISSUE #546: Design: Breadcrumb
The breadcrumb component should be styled according to [Figma](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1569%3A5671)

### Acceptance criteria

Conditions that must be satisfied to be accepted by a user, customer or other stakeholder.

### Definition of Ready

To be completed before the issue can be added to backlog.

- [ ] Acceptance criteria are described
- [ ] Any preconditions are described
- [ ] Possible dependence on external experts is mentioned
- [ ] The story must be understandable for the development team
- [ ] Any requirements for accessibility are described

### Definition of Done

To be completed before the issue can be marked as Done.

- [ ] Code exists on `main` branch
- [ ] Meets acceptance criteria
- [ ] Complies with [the standard of public code](https://standard.publiccode.net/)
- [ ] Story has been sufficiently tested
  - [ ] 90% code coverage
  - [ ] Build completed
  - [ ] Existing tests pass
- [ ] The frontend component complies with [WCAG 2.1 accessibility guidelines](https://www.w3.org/TR/WCAG21/)

paulAcato:
HTML and CSS part for the breadcrumbs based on the denhaag-link and and Utrecht breadcrumb component.

Screenshot 2021-12-06 at 11 08 38

**Closing issues**
closes #546

### DIFF
--- a/components/Breadcrumb/package.json
+++ b/components/Breadcrumb/package.json
@@ -1,0 +1,35 @@
+{
+  "name": "@gemeente-denhaag/breadcrumb",
+  "description": "A Breadcrumb component",
+  "author": "Municipality of The Hague",
+  "license": "EUPL-1.2",
+  "version": "0.2.3-alpha.63",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "exports": "./dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c ../../rollup.config.js",
+    "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nl-design-system/denhaag.git",
+    "directory": "components/Breadcrumb"
+  },
+  "bugs": "https://github.com/nl-design-system/denhaag/issues",
+  "dependencies": {
+    "@gemeente-denhaag/baseprops": "0.2.3-alpha.63",
+    "@material-ui/core": "4.12.3"
+  },
+  "devDependencies": {
+    "@gemeente-denhaag/icons": "0.2.3-alpha.63"
+  },
+  "peerDependencies": {
+    "react": "^17.0.0"
+  },
+  "gitHead": "dcf72a9b79266c1ebede35aff4a02dd9121a980f"
+}
--- a/components/Breadcrumb/src/breadcrumb.css
+++ b/components/Breadcrumb/src/breadcrumb.css
@@ -1,18 +1,10 @@
 .denhaag-breadcrumb {
-  --denhaag-breadcrumb-arrow-size: 0px;
-  --denhaag-focus-background-color: var(--denhaag-breadcrumb-link-focus-background-color);
-  --denhaag-link-background-color: var(--denhaag-breadcrumb-link-background-color);
-  --denhaag-link-color: var(--denhaag-breadcrumb-link-color);
-  --denhaag-link-focus-color: var(--denhaag-breadcrumb-link-focus-color, var(--denhaag-breadcrumb-link-color));
-  --denhaag-link-focus-text-decoration: var(--denhaag-link-text-decoration);
-  --denhaag-link-hover-color: var(--denhaag-breadcrumb-link-color);
-  --denhaag-link-hover-text-decoration: var(--denhaag-link-text-decoration);
-  --denhaag-link-visited-color: var(--denhaag-breadcrumb-link-color);
-  --denhaag-link-visited-text-decoration: var(--denhaag-link-text-decoration);
+  --denhaag-breadcrumb-spacer: 8px;
+  --denhaag-breadcrumb-background-color: #fff;
+  --denhaag-breadcrumb-transition-ease: cubic-bezier(0.55, 0, 0.1, 1);
+  --denhaag-breadcrumb-transition-speed: 375ms;
 
-  font-family: var(--denhaag-document-font-family, inherit);
-  font-size: var(--denhaag-breadcrumb-font-size);
-  text-transform: var(--denhaag-document-text-transform, inherit);
+  color: #4b4b4b;
 }
 
 .denhaag-breadcrumb__list {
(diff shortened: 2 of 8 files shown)


Robbert:
The tricky thing with not defining your design decisions in `tokens.json` but as default values for your component is, that every change to your design decisions becomes a breaking change for other users who relied on it because their (by chance) designers used the same value, like 8px padding.

paulAcato:
@Robbert , i just applied your feedback and made the component more standalone.

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child) {
+  --denhaag-breadcrumb-background-color: var(--denhaag-breadcrumb-dots-background-color, var(--denhaag-color-grey-1));
+}
+
+.denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child):hover {
+  --denhaag-breadcrumb-background-color: var(
+    --denhaag-breadcrumb-dots-hover-background-color,
+    var(--denhaag-color-grey-2)
```
bddjong, line 31: Same as the comment above

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,125 @@
...
+  padding-inline-start: 0;
+  padding-inline-end: 0;
+  position: absolute;
+  width: 1px;
+  white-space: nowrap;
+}
+
+.denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child) .denhaag-breadcrumb__text {
+  aspect-ratio: 24/16;
+  border-radius: var(--denhaag-breadcrumb-dots-border-radius, 3px);
```
bddjong, line 111: There is a default border radius token for the entire library. `--denhaag-border-radius`. That might be a better fallback value, since it allows users to define only the generic `--denhaag-border-radius` to have a consistent border radius style across all components.

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,125 @@
...
+  max-width: calc(var(--denhaag-breadcrumb-spacing, var(--denhaag-space-inline-xs)) * 3);
+  text-indent: calc(var(--denhaag-breadcrumb-spacing, var(--denhaag-space-inline-xs)) * 3);
+}
+
+.denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child) .denhaag-breadcrumb__text:before {
+  color: var(--denhaag-breadcrumb-dots-color, var(--denhaag-color-grey-3));
```
bddjong, line 117: Same as the other comments about Den Haag theme references in CSS

bddjong (changes requested):

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,125 @@
...
+[dir="rtl"] .denhaag-breadcrumb__link > .denhaag-breadcrumb__text + .denhaag-icon {
+  transform: scaleX(-1);
+}
+
+.denhaag-breadcrumb__text {
+  background: var(--denhaag-breadcrumb-background-color);
+  overflow: hidden;
+  position: relative;
+  text-overflow: ellipsis;
+  transition: background var(--denhaag-breadcrumb-transition-ease) var(--denhaag-breadcrumb-transition-speed);
```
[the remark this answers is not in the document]
paulAcato: Check, I am a fan of transitions since it makes an element just a little smoother behavior of the element since I find transitions quite hard. 

For now I'll remove the transition since it's not documented in Figma.

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.denhaag-breadcrumb__item,
+.denhaag-breadcrumb__link {
+  align-items: center;
+  display: flex;
+}
+
+.denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child) {
+  --denhaag-breadcrumb-background-color: var(--denhaag-breadcrumb-dots-background-color, var(--denhaag-color-grey-1));
```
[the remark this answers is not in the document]
paulAcato: I kinda disagree with you here. 

First of all you want the CSS variable `--denhaag-breadcrumb-dots-background-color` to be set and modified, but if it's not set you want a variable fallback value which is quite common/global. Therefor you've a fallback which is dynamic and responds to other theme colors.

components/Breadcrumb/src/breadcrumb.css
paulAcato: Ahh! Thankgod, I was looking for a more generic value, but couldn't find it. I stumbled upon the `--denhaag-card-border-radius` variant and some others which i didn't want to use. I will apply this right away.

components/Breadcrumb/src/breadcrumb.css
paulAcato: See my reaction in the topic https://github.com/nl-design-system/denhaag/pull/586#discussion_r764755985

components/Breadcrumb/src/breadcrumb.css
paulAcato: See my reaction in the topic https://github.com/nl-design-system/denhaag/pull/586#discussion_r764755985

components/Breadcrumb/src/breadcrumb.css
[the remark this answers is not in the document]
bddjong: For clarity, it's not that I think having a fallback is bad. The issue is that these fallbacks are using the Den Haag (proprietary) color palette tokens. A fallback token would only be valuable to other organisations because for Den Haag we already define the component tokens, but another organisation should not fill our color palette. Tokens like `denhaag-color-{green,red,blue,grey,orange,ocher}` have no meaning for other organisations as their color palette could have completely different colors. Which means using them as a fallback is useless. @Robbert any thoughts?

proprietary/Components/src/denhaag/breadcrumb.tokens.json
```diff
@@ -0,0 +1,51 @@
...
+        "color": {
+          "value": "{denhaag.color.grey.4.value}"
+        }
+      },
+      "dots": {
+        "background-color": {
+          "value": "{denhaag.color.grey.1.value}"
+        },
+        "border-radius": {
+          "value": "3px"
```
bddjong, line 20:
```suggestion
          "value": "{denhaag.border-radius.value}"
```

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,124 @@
+.denhaag-breadcrumb {
+  --denhaag-breadcrumb-background-color: transparent;
```
bddjong, line 2: This line has no influence on the style in my tests. I think we can remove it?

components/Breadcrumb/src/stories/breadcrumb.react.stories.mdx
```diff
@@ -0,0 +1,32 @@
...
+import pkg from "../../package.json";
+import readme from "../../README.md";
+
+<Meta
+  title="REACT Components/Navigation/Breadcrumb"
+  parameters={{
+    componentSubtitle: `${pkg.name} - ${pkg.version}`,
+    docs: { source: { type: "dynamic" } },
+    status: {
+      type: "ALPHA",
```
bddjong, line 13:
```suggestion
      type: "WORK IN PROGRESS",
```

New components should start as `WORK IN PROGRESS`. Our components generally move to Alpha once they are being used by another application.

components/Breadcrumb/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@gemeente-denhaag/breadcrumb",
+  "description": "A Breadcrumb component",
+  "author": "Municipality of The Hague",
+  "license": "EUPL-1.2",
+  "version": "0.2.3-alpha.63",
```
bddjong, line 6:
```suggestion
  "version": "0.1.0",
```

This is a new component, so it's versioning should be new too :)

components/Breadcrumb/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nl-design-system/denhaag.git",
+    "directory": "components/Breadcrumb"
+  },
+  "bugs": "https://github.com/nl-design-system/denhaag/issues",
+  "dependencies": {
+    "@gemeente-denhaag/baseprops": "0.2.3-alpha.63",
+    "@material-ui/core": "4.12.3"
```
bddjong, line 26: This component doesnt require `material-ui/core` (I hope)

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
```diff
@@ -0,0 +1,896 @@
...
+import readme from "../../README.md";
+
+<Meta
+  title="CSS Components/Navigation/Breadcrumb"
+  component={Breadcrumb}
+  parameters={{
+    componentSubtitle: `${pkg.name} - ${pkg.version}`,
+    docs: { source: { type: "dynamic" } },
+    status: {
+      type: "IN DEVELOPMENT",
```
bddjong, line 15:
```suggestion
      type: "WORK IN PROGRESS",
```

components/Breadcrumb/src/breadcrumb.css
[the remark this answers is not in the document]
paulAcato: @bddjong , Not sure who, but I was told that when this went to the nl-design repository the prefixes would be set to `--nl-design` instead of `--denhaag`. When that happens, doesn't that tackle the whole issue?

proprietary/Components/src/denhaag/breadcrumb.tokens.json
paulAcato: Not sure what yo mean by that.

components/Breadcrumb/src/breadcrumb.css
paulAcato: The dot's  overrule this css property, by default it's transparent. Without it Safari bugged at my macbook. Rendered "null" in the value box of the editor.

components/Breadcrumb/src/stories/breadcrumb.react.stories.mdx
paulAcato: Copied form the Buttons or Link, forgot to change this.

components/Breadcrumb/package.json
paulAcato: Copied form the Buttons or Link, forgot to change this (also didn't knew we applied a version before it's fully completed?).

components/Breadcrumb/package.json
paulAcato: Removed it, it indeed doesn't require material-ui

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
paulAcato: edited

components/Breadcrumb/package.json
bddjong: The CI will bump it to something like `0.1.1-alpha.0` :D

components/Breadcrumb/src/breadcrumb.css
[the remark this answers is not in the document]
Robbert: @paulAcato I agree there is no need for the `--denhaag-color-grey-1` fallback: all components should only contain "semantic" design tokens such as component level tokens and common tokens regarding common patterns such as focus color. Any design decisions should strictly be separated from the component code and they should be documented in Style Dictionary, so they remain in sync with Figma where we also (intend to) use `tokens.json` files.

Having these design tokens makes the code a bit less "forkable" too, because now you not only need to rename the vendor prefix to your own prefix, but also clean up these design tokens that are basically dead code for the rest of the world.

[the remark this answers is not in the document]
paulAcato: I'll replace my fallback values to hard values from the design.

components/Breadcrumb/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nl-design-system/denhaag.git",
+    "directory": "components/Breadcrumb"
+  },
+  "bugs": "https://github.com/nl-design-system/denhaag/issues",
+  "dependencies": {
+    "@gemeente-denhaag/baseprops": "0.2.3-alpha.63"
```
bddjong, line 25: This component does not use the `baseprops`

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,121 @@
...
+  margin-block-end: -1px;
+  overflow: hidden;
+  padding-block-start: 0;
+  padding-block-end: 0;
+  padding-inline-start: 0;
+  padding-inline-end: 0;
+  position: absolute;
+  width: 1px;
+  white-space: nowrap;
+}
```
bddjong, line 103: Similar to in Pagination, I wonder if we can avoid this class by proper usage of aria-attributes.

components/Breadcrumb/src/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import BaseProps from '@gemeente-denhaag/baseprops';
+import './breadcrumb.css';
+
+export interface BreadcrumbProps extends BaseProps {}
+
```
bddjong, line 6: I know this PR isn't focussed on React, but to avoid confusion when the React component gets implemented. I think it's better if we can remove any usages of BaseProps. Also, the BreadcrumbProps interface is unused for now, so we can also remove that.

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
```diff
@@ -0,0 +1,896 @@
...
+        <li
+          class="denhaag-breadcrumb__item"
+          itemscope
+          itemtype="https://schema.org/ListItem"
+          itemprop="itemListElement"
+        >
+          <a
+            class="denhaag-breadcrumb__link denhaag-breadcrumb__link--current"
+            href="https://example.com/a/b/"
+            aria-current="location"
```
bddjong, line 102: According to [WAI ARIA](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html), this should be `aria-current="page"`

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
```diff
@@ -0,0 +1,896 @@
...
+<DesignTokenDocBlock categoryName="Breadcrumb" viewType="table" />
+
+## Stories
+
+### Default
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="Default">
+    <nav class="denhaag-breadcrumb">
```
bddjong, line 33: [WAI ARIA](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html) practices suggest an `aria-label="Breadcrumb"` on the `<nav>` element.

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
bddjong: Applies to the other stories as well

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
bddjong: Applies to the other stories as well

bddjong (changes requested):

components/Breadcrumb/src/breadcrumb.css
```diff
@@ -0,0 +1,121 @@
+.denhaag-breadcrumb {
+  --denhaag-breadcrumb-background-color: transparent;
+  color: var(--denhaag-breadcrumb-color, #4b4b4b);
```
bddjong, line 3: In a few places there are still hardcoded colors. There are design choices and should in my opinion be discarded. It's OK to expect other organisations to define certain tokens.

bddjong (changes requested):

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
paulAcato: Will do

components/Breadcrumb/src/breadcrumb.css
paulAcato: Replaced by inherit over the whole document. Since otherwise we overwrite a value with an optional one, which can cause issues.

components/Breadcrumb/src/stories/breadcrumb.bem.stories.mdx
paulAcato: Will do

components/Breadcrumb/src/index.tsx
paulAcato: Done

components/Breadcrumb/src/breadcrumb.css
paulAcato: Replaced with Aria

components/Breadcrumb/package.json
paulAcato: removed it.

bddjong (changes requested):
Looking gooood! Last comment I only just noticed, we can add this components CSS to the `components-css` package. (Import the css file in `packages/components-css/src/index.tsx` so it's exported in that package)

PR #602, opened 2021-12-10
[DRAFT] Table component

bddjong:
In progress


components/Table/src/TableColGroup.tsx
```diff
@@ -0,0 +1,17 @@
...
+import clsx from 'clsx';
+
+export type TableColGroupProps = ColgroupHTMLAttributes<HTMLTableColElement>;
+
+/**
+ * TableColGroup component based on the <colgroup> element.
+ * Defines a group of columns within a table.
+ */
+export const TableColGroup: React.FC<TableColGroupProps> = ({ className, ...props }: TableColGroupProps) => {
+  const rootClassNames = clsx('denhaag-table__colgroup', className);
```
mjcarsjens, line 11: Inconsistent use of casing wrt the classname here.
Usually the places where we case a letter in a component name would lead to a `-` in the classname.

components/Table/src/Table.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './index.css';
```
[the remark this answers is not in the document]
Robbert: The only downside that comes to mind is `Content-Security-Policy` / `unsafe-inline`

components/Table/src/stories/react.stories.mdx
```diff
@@ -0,0 +1,57 @@
...
+          <TableCell></TableCell>
+          <TableHeader scope="col" className="heman">
+            He-Man
+          </TableHeader>
+          <TableHeader scope="col" className="skeletor">
+            Skeletor
+          </TableHeader>
+        </TableRow>
+        <TableRow>
+          <th scope="row">Role</th>
```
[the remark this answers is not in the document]
Robbert: Maybe we can even have RowHeader and ColumnHeader.

components/Table/src/stories/react.stories.mdx
```diff
@@ -0,0 +1,211 @@
...
+            <TableRow tabIndex="-1">
+              <TableCell></TableCell>
+              <TableHeader className="heman">He-Man</TableHeader>
+              <TableHeader className="skeletor">Skeletor</TableHeader>
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            <TableRow>
+              <TableHeader scope="row">Role</TableHeader>
+              <TableCell data-heading="He-Man">Hero</TableCell>
```
mjcarsjens, line 79: What's the exact reasoning behind the `data-heading` attributes?

bddjong: @paulAcato correct me if I'm missing something. From what I understood, the `data-heading` attribute is used in the CSS to give some flexibility in creating a different layout for mobile. On mobile display sizes, a title is placed before each cell using the `attr()` css function. 

The advantage is that it allows us to support a completely different table layout for mobile (vertical, as per [our design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=5213%3A16235)), without creating a whole new component. 

A few disadvantages I can think of:
- it forces you to add the heading attribute to every cell, this could be very tedious for tables that are hand-made;
- there's a risk of inconsistent titles if a developer forgets to update the `data-heading` when the table header is changed;
- we're introducing complexity in using the BEM classes (although it's optional);
- semantically, a vertical table is very similar to a [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl). It might be better to just create a description list component and leave it to developers to swap them accordingly.

bddjong: @Robbert any thoughts on this?

Robbert: @bddjong I think during the refinements I made similar points, I still have to see the first accessible responsive table. My preference was too to provide an alternative definition list for specific cases. If I recall correctly, optimism about the feasibility and challenges in WordPress were reasons to try this approach first. But I'm not convinced yet :)

Issue #428, opened 2021-09-22
Keyboard a11y not being tested properly


mjcarsjens:
The keyboard accessibility is not being tested properly. When calling `cy.get('...').type('{enter}')` what it actually does is the following:
click the element to receive focus, and then type enter. The click being registered is actually the focus click instead of the enter typing. When running `cy.get('...').focus().type('{enter}')` you will see that the `onClick` is actually not called at all. 

Weirdly enough, this is the case for natively clickable element (e.g. `a` and `button` tags, but when a non-natively-clickable element is made focusable (with `tabindex="0"`) and clickable (e.g. our `Listitem` component makes a `li` tag clickable) the `cy.get('...').focus().type('{enter'})` does in fact call the `onClick` function. Need to find a way to make keyboard behavior properly testable. 

Relevant cypress issue describing typing `enter` does not in fact call a `onClick`: https://github.com/cypress-io/cypress/issues/8267

bddjong:
This plugin could solve the issue: https://github.com/dmtrKovalenko/cypress-real-events#readme

Using `.realClick()`

PR #603, opened 2021-12-10
DRAFT: Feature/#479 datepicker

### LINKED ISSUE #479: OpenForms Flatpickr style overwrite to support design tokens
### User story

As a developer / organisation, I want to be able to style the date picker component in OpenForms with design tokens to match my organisation theme. 

### Description

OpenForms uses [FlatPickr](https://flatpickr.js.org/) for its date picker input. This means the styling is encapsulated in a node module, instead of in the open forms source. 

### Describe alternatives you've considered

Overwriting the Flatpickr CSS to use css variables referring to design tokens with fallback values. This however leads to another coupling with a third party library that would need to be maintained. 

### Additional context

image

mjcarsjens:
Datepicker component

- Added comment in figma for circular error icon (https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn?node-id=3497:14242#134280692)

### A11y
- Implemented keyboard navigation as described in: [https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html)

**_Note_**: created an issue for this specific piece of documentation since the example implementation here has some faulty behaviour with regards to keyboard navigation, specifically `Page Up` and `Page Down` navigation on the first and last week of months ([issue](https://github.com/w3c/aria-practices/issues/2170)).

- Added aria, role and abbr attributes to components as described in the same documentation

**_Note_**: the documentation states that the button to the corresponding date should have `aria-selected="true"`. However, this is in fact not valid as `aria-selected` is not allowed on elements with role `button` (which is implicit on the `button` HTML tag): [issue](https://github.com/w3c/aria-practices/issues/2171).

**Closing issues**
closes #479


components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+  isSameMonth,
+  setDate,
+} from 'date-fns';
+import { nl } from './locale';
+
+import BaseProps from '@gemeente-denhaag/baseprops';
+
+import './datepicker.css';
+
+export interface DatepickerProps extends BaseProps {
```
bddjong, line 21: I wonder if we want to use ` extends InputHTMLAttributes<HTMLInputElement>` instead. This will make sure we can support all the expected attributes on an input, and pass them to the native input within our wrapper.

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+        >
+          {date.getDate()}
+        </button>
+      );
+    }
+    return '';
+  };
+
+  return (
+    <div className={rootStyles} ref={componentRef}>
```
bddjong, line 105: Shouldn't the ref for this component be on the `<input>`? Unless I'm misunderstanding refs.

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+      setState({
+        ...state,
+        opened: false,
+      });
+    }
+  };
+
+  React.useEffect(() => {
+    document.addEventListener('click', outsideClickListener);
+    return () => {
```
bddjong, line 82: I prefer returning a named function as it's much more obvious what's happening when it's called `cleanup`, but that's a preference thing. (Caused by the fact that I had no idea what returning a function in `useEffect` would do)

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+        onClick={(event) => {
+          event.preventDefault();
+          setState({ ...state, opened: true });
+        }}
+        onKeyDown={(event) => {
+          if (event.key === 'Enter') {
+            event.preventDefault();
+            setState({ ...state, opened: true });
+          }
+        }}
```
bddjong, line 135: Pressing `Space` on Chrome (Win 11) opens the browser datepicker. 
image

bddjong (changes requested):

components/Datepicker/src/index.tsx
mjcarsjens: Changed :)

components/Datepicker/src/index.tsx
mjcarsjens: Fair enough, nice improvement to the readability. Now returns a named `outsideClickListenerCleanup` function.

components/Datepicker/src/index.tsx
mjcarsjens: Fixed with the implementation of all keyboard functionality.

components/Datepicker/src/index.tsx
mjcarsjens: The `componentRef` is used for the `outsideClickListener` so this does have to be on the outer `div` instead of the `input` element. Made a few changes though, exposed `ref` so the user can also add a `ref` of their own (replacing `componentRef`) and added a `inputRef` prop which the user can use to supply a `ref` to the HTML `input` element.

components/Datepicker/package.json
```diff
@@ -22,9 +22,11 @@
     "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
   },
   "dependencies": {
-    "@material-ui/pickers": "3.3.10"
+    "@gemeente-denhaag/baseprops": "0.2.3-alpha.71",
```
bddjong, line 25: Baseprops arent needed here anymore 🎉

components/Datepicker/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "@gemeente-denhaag/pickersutilsprovider",
```
bddjong, line 2: Do these old picker packages need to be deprecated from npm?

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,366 @@
...
+        </div>
+        <table
+          className="denhaag-datepicker__calendar-table"
+          role="grid"
+          aria-labelledby="denhaag-datepicker-calendar-month"
+        >
+          <tbody>
+            <tr>
+              {Array.from(Array(7)).map((_, i) => (
+                <th
```
bddjong, line 344: I think we should give the `<th>` a `scope=col` attribute here to indicate its a column header

bddjong (changes requested):

components/Datepicker/package.json
mjcarsjens: Already did that 👍🏼

components/Datepicker/src/datepicker.css
```diff
@@ -0,0 +1,220 @@
...
+  color: var(--denhaag-icon-button-hover-color);
+}
+
+.denhaag-datepicker__calendar-navigation:focus {
+  color: var(--denhaag-icon-button-focus-color);
+  outline-color: var(--denhaag-icon-button-focus-outline-color);
+  outline-width: var(--denhaag-icon-button-focus-outline-width);
+  outline-style: var(--denhaag-icon-button-focus-outline-style);
+  outline-offset: var(--denhaag-icon-button-focus-outline-offset);
+}
```
bddjong, line 157: This focus state isn't documented in Figma. I added a [comment in figma](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn?node-id=1681:7756#135259358).

Also, wouldn't it make sense to use the IconButton component for this?

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,400 @@
...
+   */
+  inputRef?: RefObject<HTMLInputElement>;
+  /**
+   * Whether the input field is in a error state.
+   */
+  error?: boolean;
+  /**
+   * Whether the input field is disabled.
+   */
+  disabled?: boolean;
```
bddjong, line 52: The `disabled` prop should already be supplied by the [InputHTMLAttributes](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) interface

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,400 @@
...
+  const currentButtonRef = React.useRef<HTMLButtonElement>(null);
+
+  const outsideClickListener = (e: Event) => {
+    if (!ref?.current?.contains(e.target as Node) && state.opened) {
+      setState({
+        ...state,
+        opened: false,
+      });
+    }
+  };
```
bddjong, line 126: This code doesn't seem to work. I'm unable to close the dialog by clicking outside of the element. (Tested in both Chrome and Edge)

components/Datepicker/src/datepicker.css
bddjong: Figma now documents the IconButton focus, which should also be rectangular. https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1681%3A7756

bddjong (changes requested):
Looks good. Control wise I noticed a few difficulties in usage. 
- The dialog is hard to close (perhaps we should have a close icon?) 
- Selecting 1st/2nd/3rd of December 2021 and pressing page down will move to another date in December, instead of going to the next month. This seems to occur on months that cover 5 weeks.

components/Datepicker/tsconfig.json
```diff
@@ -9,7 +9,7 @@
   "exclude": ["dist", "src/**/*.stories.tsx"],
   "references": [
     {
-      "path": "../PickersUtilsProvider"
+      "path": "../BaseProps"
```
bddjong, line 12: We can also remove BaseProps here

components/Datepicker/src/index.tsx
```diff
@@ -0,0 +1,400 @@
...
+        aria-label={openCalendarLabel}
+        onClick={() => {
+          setState({ ...state, opened: !state.opened });
+        }}
+        tabIndex={props.disabled ? -1 : 0}
+        ref={openButtonRef}
+      >
+        <CalendarIcon />
+      </button>
+      <input
```
bddjong, line 297: On safari, the date separators in the placeholder are darker, while on chrome they're the entire placeholer is grey. 
**Safari:** 
image
**Chrome**
image

bddjong:
I also noticed on the docs tab, the stories canvases are too small to display the dialog

image

components/Datepicker/src/index.tsx
mjcarsjens: Fixed!

components/Datepicker/src/datepicker.css
mjcarsjens: Squared it is :(

mjcarsjens: I tried using the `IconButton` here but I ran into some issues with the `onKeyDown` not working for our own  `IconButton` component. Don't want to spend too much time now debugging this, but nice to keep on our radar that when we remove MUI from the `IconButton` we should use it here.

components/Datepicker/src/index.tsx
mjcarsjens: Fixed!

components/Datepicker/src/index.tsx
mjcarsjens: Should work now :)

components/Datepicker/tsconfig.json
mjcarsjens: Done :)

mjcarsjens:
> I also noticed on the docs tab, the stories canvases are too small to display the dialog
> 
> image

fixed :)

mjcarsjens:
> Looks good. Control wise I noticed a few difficulties in usage.
> 
>     * The dialog is hard to close (perhaps we should have a close icon?)
> 
>     * Selecting 1st/2nd/3rd of December 2021 and pressing page down will move to another date in December, instead of going to the next month. This seems to occur on months that cover 5 weeks.

Also fixed the page up/page down bug on certain days

components/Datepicker/src/index.tsx
mjcarsjens: Reworked the placeholders to work in all browsers, so this does not occur anymore

mjcarsjens:
Added compontent tests, with one small caveat: keyboard navigation is not currently being tested as this isn't supported properly by cypress by default (#428)

bddjong (approved):
My new favorite component :D

Keyboard navigation works great. Screen readers are still a bit iffy. We should have a follow-up issue to investigate the best approach in making it consistent and user friendly for users relying on screen readers. Ubuntu screen reader announces a lot of unnecessary information (random dates in the table)

PR #619, opened 2021-12-17
feat(html-divider): setup stories for html divider, and extend css wi…

### LINKED ISSUE #618: HTML divider
Add story (HTML) and CSS for the divider.

paulAcato:
Made HTML story for the Divider, also extended the styles minimal with the material-design styles.

closes #618

### DIFF
--- a/components/Divider/src/divider.css
+++ b/components/Divider/src/divider.css
@@ -1,8 +1,10 @@
 .denhaag-divider {
-  height: var(--denhaag-divider-border-width);
   background: var(--denhaag-divider-border-color);
+  border: none;
+  height: var(--denhaag-divider-border-width);
   margin-block-start: var(--denhaag-divider-margin-block-start);
   margin-block-end: var(--denhaag-divider-margin-block-end);
+  flex-shrink: 0;
 }
 
 .denhaag-divider--vertical {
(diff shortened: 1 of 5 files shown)


components/Divider/src/stories/bem.stories.mdx
```diff
@@ -0,0 +1,73 @@
...
+  </Story>
+  {/* eslint-enable react/no-unknown-property */}
+</Canvas>
+
+### Vertical
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="Vertical">
+    <div style={{ display: "flex" }}>
```
robbieatacato, line 42: @paulAcato I think it's cleaner to assign a css class to the `<div>` instead of inline css.

components/Divider/src/divider.css
```diff
@@ -1,8 +1,10 @@
 .denhaag-divider {
-  height: var(--denhaag-divider-border-width);
   background: var(--denhaag-divider-border-color);
+  border: none;
```
robbieatacato, line 3: @paulAcato I prefer: border: 0
Because it's shorter.
But both are possible.

robbieatacato (approved):

components/Divider/src/stories/bem.stories.mdx
bddjong: Although I would usually agree, since this is just a single style for storybook demo purposes only, I don't mind an inline style to set it to flex.

bddjong (approved):

bddjong (approved):

bddjong (changes requested):
There's still a reference to `divider.css`, I'm guessing in the components-css package.

robbieatacato:
@bddjong I also changed the css file name in the 'components-css' directory.

bddjong (approved):

PR #630, opened 2021-12-23
Feature/#610 blockquote

### LINKED ISSUE #610: Design: Text block quote
### User story

A clear and concise description of what the problem is. Eg: As an [actor], I [want|must] [action] so that [achievement].

### Acceptance criteria

Conditions that must be satisfied to be accepted by a user, customer or other stakeholder.

### Definition of Ready

To be completed before the issue can be added to backlog.

- [ ] Acceptance criteria are described
- [ ] Any preconditions are described
- [ ] Possible dependence on external experts is mentioned
- [ ] The story must be understandable for the development team
- [ ] Any requirements for accessibility are described

### Definition of Done

To be completed before the issue can be marked as Done.

- [ ] Code exists on `main` branch
- [ ] Meets acceptance criteria
- [ ] Complies with [the standard of public code](https://standard.publiccode.net/)
- [ ] Story has been sufficiently tested
  - [ ] 90% code coverage
  - [ ] Build completed
  - [ ] Existing tests pass
- [ ] The frontend component complies with [WCAG 2.1 accessibility guidelines](https://www.w3.org/TR/WCAG21/)

robbieatacato:
## Solving:
https://github.com/nl-design-system/denhaag/issues/610

## Purpose:
- visualize quoted text
- 2 variants: with and without author name

closes #610


components/Blockquote/src/index.css
```diff
@@ -0,0 +1,28 @@
+.denhaag-blockquote {
+  border-inline-start: 4px solid var(--denhaag-color-green-1, #d5e7d4);
```
bddjong, line 2: Throughout the library we avoid using tokens from the `Proprietary` package in CSS. Instead, tokens used in CSS should usually be component specific. `--denhaag-blockquote-border-color`. The CSS should by default contain no proprietary styles, as another organisation should be able to use it and have to make their own design decisions, rather than using the Den Haag style as default.

components/Blockquote/src/index.css
```diff
@@ -0,0 +1,28 @@
...
+  padding-block-start: 1rem;
+  padding-block-end: 1rem;
+  padding-inline-start: 1rem;
+  padding-inline-end: 1rem;
+}
+
+.denhaag-blockquote__content p {
+  color: var(--denhaag-color-green-4, #1d6b34);
```
bddjong, line 10: Same as the comment above about tokens.

components/Blockquote/src/index.css
```diff
@@ -0,0 +1,28 @@
...
+  font-family: var(--denhaag-typography-sans-serif-alternate-font-family, inherit);
+  font-size: var(--denhaag-typography-scale-lg, 1.25rem);
+  font-weight: 700;
+  line-height: 1.3;
+  margin-block-start: 0;
+  margin-block-end: 0;
+}
+
+.denhaag-blockquote__content .denhaag-blockquote__attribution {
+  color: var(--denhaag-color-grey-4, #4b4b4b);
```
bddjong, line 20: Ditto.

bddjong (changes requested):

components/Blockquote/src/index.css
paulAcato: Additional to @bddjong comment, i would suggest to split this onto three CSS variables, otherwise we're making a design choice for the person/client using this component.

My suggestions are:
1. `--denhaag-blockquote-border-color`
2. `--denhaag-blockquote-border-width`
3. `--denhaag-blockquote-border-style`

components/Blockquote/src/stories/bem.stories.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+<DesignTokenDocBlock categoryName="Blockquote" viewType="table" />
+
+## Stories
+
+### Default
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="Default">
```
paulAcato, line 30: If I check the component on MDN https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote I prefer their HTML semantic, it's more cleaner since there's no (unnecessary) `<div>` with it.

components/Blockquote/src/index.css
```diff
@@ -0,0 +1,28 @@
...
+  color: var(--denhaag-color-green-4, #1d6b34);
+  font-family: var(--denhaag-typography-sans-serif-alternate-font-family, inherit);
+  font-size: var(--denhaag-typography-scale-lg, 1.25rem);
+  font-weight: 700;
+  line-height: 1.3;
+  margin-block-start: 0;
+  margin-block-end: 0;
+}
+
+.denhaag-blockquote__content .denhaag-blockquote__attribution {
```
paulAcato, line 19: This CSS selector doesn't need to be scoped, since it's prefixed with `.denhaag-blockquote__`. Also i would suggest using the `citer` or `quoter` since it's shorter and says more about the element.

paulAcato (changes requested):

components/Blockquote/src/stories/bem.stories.mdx
robbieatacato: @paulAcato This is based on the Utrecht way https://nl-design-system.github.io/utrecht/?path=/docs/css-component-blockquote--blockquote

robbieatacato:
@bddjong ik heb de bovenstaande aanpassingen doorgevoerd in deze commit: https://github.com/nl-design-system/denhaag/commit/20a9564402dadf08e08ecf7002689cb27ae8cde6

N.B. Ik heb ook nog een extra commit gedaan daarna, om de build te fixen.
Alleen dit zat in de Note component (niet in de Blockquote), zie commit: https://github.com/nl-design-system/denhaag/commit/0e03e8fa2002c36ccf6b5e47e00722bec2c3b844
Is dit ok?

components/Blockquote/src/stories/bem.stories.mdx
bddjong: I agree, if there's a semantic tag we should use it

Issue #576, opened 2021-12-02
Design: Accordion


paulAcato:
### User story

Create a accordion component

robbieatacato:
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1585%3A0#

PR #637, opened 2021-12-23
feat(accordion): add bem story, togglescript, css + tokens

robbieatacato:
Solving:
https://github.com/nl-design-system/denhaag/issues/576

Purpose:
- adding documentation in bem stories
- visualizing an accordion component in different states


components/Accordion/src/index.css
```diff
@@ -0,0 +1,124 @@
+/* accordion general styling */
```
bddjong, line 1: I don't think we need these comments

components/Accordion/src/index.tsx
```diff
@@ -1,4 +1,6 @@
 import Accordion from './Accordion';
+import './index.css';
+import './stories.js';
```
bddjong, line 3: This would probably result in stories.js being added to the build and NPM packages. This script is only for use in storybook right? In that case we should probably import it in preview.js or another storybook-only file.

proprietary/Components/src/denhaag/accordion.tokens.json
```diff
@@ -0,0 +1,22 @@
+{
+  "denhaag": {
+    "accordion": {
+      "border": { "value": "1px solid {denhaag.color.grey.2.value}" },
```
bddjong, line 4: `.value` can be omitted

bddjong (changes requested):

robbieatacato:
@bddjong I made changes after the request above.

bddjong (approved):

PR #643, opened 2022-01-04
refactor: omit .value in token reference

Robbert:
Let's get rid of the most annoying thing of the design token JSONs: the `.value` at the end of references. Omitting `.value` is supported since style-dictionary@3.1.0.

https://github.com/amzn/style-dictionary/pull/746

Tested it by diffing the `dist/index.css` before and after: no difference.


bddjong (approved):
Thats awesome! One of the most common reasons for weird errors is forgetting the `value` :')

Issue #607, opened 2021-12-14
Design: Typography


jeroenkel:
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=3497%3A14122

PR #700, opened 2022-01-13
feat(typography): add stories, modify css

robbieatacato:
## Solve
https://github.com/nl-design-system/denhaag/issues/607

## Purpose
Heading and paragraph:
- html
- css

#closes https://github.com/nl-design-system/denhaag/issues/607


robbieatacato:
@bddjong I fixed the merge conflict with the main branch.
Can you check this pr, this one is essential to multiple components.

bddjong:
We're not sure where the detail text would be used. I'd like to avoid adding unused components to the library. Blocking this PR for now until there's a use case documented in the issue.

robbieatacato:
> We're not sure where the detail text would be used. I'd like to avoid adding unused components to the library. Blocking this PR for now until there's a use case documented in the issue. There's a single use in the posttype label, but I don't think a single text variation should be a reason to create a separate typography component. @mjcarsjens @Robbert

Any progression?

components/Typography/src/paragraph.css
```diff
@@ -20,3 +20,17 @@
...
+  text-transform: uppercase;
+}
+
+.denhaag-paragraph--detail-small {
+  font-size: var(--denhaag-typography-scale-s-font-size, 0.75rem);
+}
+
+.denhaag-paragraph--detail-extra-small {
+  font-size: var(--denhaag-typography-scale-xs-font-size, 0.625rem);
+}
```
bddjong, line 36: These variations can probably be removed, using a default `--detail` font-size of 12px (.75rem)

bddjong (changes requested):
The discussion has been resolved for now. See comment on fix

components/Typography/src/paragraph.css
robbieatacato: I refactored it, removed the extra small detail story + re-used the posttype-label tokens.

bddjong (approved):

PR #638, opened 2021-12-23
Feature/633 data list

paulAcato:
HTML and CSS BEM story of the DATA list.

Files:
- every file in '/components/DescriptionList/ ' directory
- 'proprietary/Components/src/denhaag/description.tokens.json'
- 'packages/components-css/src/index.tsx'

closes #633

PR #698, opened 2022-01-13
feat(descriptionlist): add story, css, tokens

robbieatacato:
## Solves
https://github.com/nl-design-system/denhaag/pull/638

## Purpose
Description list:
- html
- css
- tokens

closes https://github.com/nl-design-system/denhaag/pull/638

PR #738, opened 2022-02-08
feat(descriptionlist): add comp + solve merge conflict main branch

robbieatacato:
Solve:
https://github.com/nl-design-system/denhaag/pull/698

Purpose:

- re-add component descriptionlist
- solve merge conflict with the main branch

closes #638

### DIFF
--- a/components/DescriptionList/README.md
+++ b/components/DescriptionList/README.md
@@ -1,0 +1,90 @@
+# Data list
+
+Data lists organize and display information into rows.
+
+## When to use
+
+Use a data list for displaying data that is intended to be easily scanned and compared.
+
+## Alternatives and related components
+
+- Use lists to create an actionable list of related items that link to detail pages, such as a list of customers.
+- Use tables to display information from a data set into rows and columns.
+
+## Anatomy
+
+The data list consists of:
+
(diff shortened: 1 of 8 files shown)


bddjong:
This PR is [missing changes](https://github.com/nl-design-system/denhaag/pull/698/files#diff-b2faf6b2e5aff45a12355ffc8a1893b65d57d68daf4e8080ab5cb5e1e206cf87R4-R7) from your original PR causing the build to fail. If you want to see why a check is failing, you can read the CI output by clicking on details next to the failed check. 

If we can we should really avoid making a new PR. It's hard to keep track of the discussion, and (as evident by the failed build) it's easy to forget some changes. Not to mention some valuable feedback might get lost.

components/DescriptionList/README.md
```diff
@@ -0,0 +1,90 @@
+# Data list
```
bddjong, line 1: Didn't notice this before, but we should probably just refer to it as a description list everywhere. Now the docs are still calling it a data list. If we change this, we should leave a comment behind in FIgma about the naming so the Figma is updated as well.

bddjong (approved):

PR #743, opened 2022-02-11
Feature/#59 textfield

### LINKED ISSUE #59: Design: Text field
The TextField component needs to be styled according to the [Figma design](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1%3A1653).

## Acceptance criteria
**Variants implemented**
- [x] Single line
- [ ] Multi line

**States implemented**
- [x] Normal
- [x] Hover
- [x] Disabled
- [x] Error
- [x] Focus

**Components**
- [ ] Optional placeholder text
- [ ] Optional helper text
- [ ] Required label
- [ ] Optional error text
- [ ] For the multi-line variant an optional character counter can be added

**Icon support**
- [x] An optional trailing icon can be added in the text filed
- [x] An optional error icon can be added in the text field (when a trailing icon is shown, the error icon will appear before it)

For this we have to use the MUI `InputAdornment` component

**Requirements**
- [ ] The optional character counter should be supplied a max-length and update the current amount of characters when typing.
- [x] [Alignment](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System-Gemeente-Den-Haag?node-id=1%3A1653)  requirements are met

bddjong:
**Changes**
- [x] Use Utrecht BEM classes for textfield 
- [x] Add error state
- [x] Add icon support in textfield
- [x] Add stories
- [x] Update Utrecht `form-input` tokens to match Den Haag theme
- [x] Add `circle-warning-icon`

`form-control` and `form-control-label` components will be added as a separate component outside of this PR.

**Closing issues**
Closes #59


bddjong:
Focus still broken on Safari 🎉

components/TextField/src/index.scss
```diff
@@ -0,0 +1,86 @@
...
+.denhaag-textfield:focus-within::after {
+  border: var(--denhaag-focus-border);
+  border-radius: var(--denhaag-border-radius);
+  content: "";
+  display: block;
+  position: absolute;
+  top: -2px;
+  right: -2px;
+  bottom: -2px;
+  left: -2px;
```
mjcarsjens, line 19: @bddjong 

Just thought of something (bit late, i know), but this only really works in our specific scenario. To fix it with a different set of design tokens this should be: 

`calc(-1 * var(--denhaag-focus-border-width))` instead of a plain old `-2px`.

PR #749, opened 2022-02-18
build: stylelint scss config

bddjong:
Stylelint was throwing error on usage of `@extend`. Added scss config to stylelint.

Copied new config rules from [utrecht config](https://github.com/nl-design-system/utrecht/blob/main/.stylelintrc.json)

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,9 +1,17 @@
 {
+  "extends": ["stylelint-config-standard-scss"],
   "rules": {
+    "scss/at-rule-no-unknown": true,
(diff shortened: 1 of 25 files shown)


mjcarsjens (approved):
Looks good. Quick question: do we even want to support $ variables?

bddjong:
> Looks good. Quick question: do we even want to support $ variables?

I can imagine there might be use cases for sass variables. (Usage in media queries?). That also keeps a nice distinction between design tokens as css variables, and functional variables as sass variables

PR #766, opened 2022-02-23
feat(listelement): add list element html, css, tokens

### LINKED ISSUE #754: Lists
- unordered list
- ordered list
- alphabetical list

Design: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=8105%3A22880

robbieatacato:
Solve:
https://github.com/nl-design-system/denhaag/issues/754

Purpose:
- html + css for unordered list
- html + css for ordered list
- html + css for lower alpha list
- including tokens

closes #754

### DIFF
--- a/components/ListElement/package.json
+++ b/components/ListElement/package.json
@@ -1,0 +1,27 @@
+{
+  "name": "@gemeente-denhaag/listelement",
+  "version": "0.1.0",
+  "description": "A list element component",
+  "bugs": "https://github.com/nl-design-system/denhaag/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nl-design-system/denhaag.git",
+    "directory": "components/Typograhy"
+  },
+  "license": "EUPL-1.2",
+  "author": "Municipality of The Hague",
+  "exports": "./dist/index.js",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c ../../rollup.config.js",
+    "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
+  },
+  "peerDependencies": {
+    "react": "^17.0.0"
+  }
+}
--- a/components/ListElement/src/index.scss
+++ b/components/ListElement/src/index.scss
@@ -1,0 +1,17 @@
+.denhaag-list-element {
+  color: var(--denhaag-list-element-color);
+  font-family: var(--denhaag-list-element-font-family);
+  font-size: var(--denhaag-list-element-font-size);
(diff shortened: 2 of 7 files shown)


components/ListElement/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@gemeente-denhaag/listelement",
```
paulAcato, line 2: "listelement" shouldn't this just be "list"? Currently you're suggestion other elements are no elements xD

@bddjong any thoughts?

components/ListElement/src/index.scss
```diff
@@ -0,0 +1,17 @@
+.denhaag-list-element {
```
paulAcato, line 1: Same here, the CSS can be shorter when `-element` isn't added. In my opinion it's unnecessary.

paulAcato (changes requested):

components/ListElement/package.json
robbieatacato: 'list' already exists

components/ListElement/src/index.scss
robbieatacato: 'list' already exists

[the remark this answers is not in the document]
robbieatacato: maybe in the future there will be ordered list specifik styling

components/ListElement/package.json
bddjong: I would suggest following the utrecht approach. Separating the unordered list and ordered list into separate components, and keeping the naming as close to the element names as possible.

`denhaag/unordered-list` and `denhaag/ordered-list`

https://github.com/nl-design-system/utrecht/tree/main/components/unordered-list
https://github.com/nl-design-system/utrecht/tree/main/components/ordered-list

This will also allow us to use `@extend .utrecht-unordered-list` for example, to use their CSS and design tokens.

components/ListElement/src/index.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin-inline-start: var(--denhaag-list-element-margin-inline-start);
+  padding-inline-start: var(--denhaag-list-element-padding-inline-start);
+
+  &--unordered li::marker {
+    font-size: var(--denhaag-list-element-unordered-font-size);
+  }
+
+  &--lower-alpha {
+    list-style-type: var(--denhaag-list-element-lower-alpha-list-style-type);
+  }
```
bddjong, line 16: We've decided not to use nesting selectors (or the parent selector) features of scss (for now). One of the benefits being that without nesting, you can easily search your IDE for BEM class names. @Robbert any thoughts on this? I noticed Utrecht doesn't use it either. Are there more reasons to avoid using this syntax?

bddjong (changes requested):

PR #776, opened 2022-02-25
feat(mui-removal): reworked List component

### LINKED ISSUE #345: (List): Remove MUI from List (fix bugs)
## Bug

The [list component](http://dhreactstorybook-staging.azurewebsites.net/?path=/story/components-data-display-list--list-with-icons) has inconsistent tab behavior across browsers. In Chrome, a secondary action list item can only have its action focussed, but in Safari, the entire list item is skipped when tabbing.

Additionally, pressing a list item with a secondary action gives the entire list item a focus effect.

### Expected behavior
According to Figma, only the action icon should be focusable if present. The list item (row) itself should not be focusable. When using tab to navigate through actions on a webpage, the icon should focus.

### Current behavior
The entire list item is focusable with a click on it. 
When tabbbing:
- In Google Chrome, the action icon takes focus
- In Safari, the entire list item row is skipped. Clicking on the list item row gives it the incorrect focus mentioned earlier. Clicking on the action icon however has no effect, even though it should focus the icon (as it does in Chrome).

## Suggested fix
Removing Material UI should resolve inconsistent focus behaviour. Also look into which HTML tags should be used for accessibility.

## Acceptance criteria 
- [ ] List component works without Material UI
- [ ] Focus behaviour is consistent
- [ ] Non navigation list items should not be navigateable with tab.

### LINKED ISSUE #402: List Issues
- [ ] a11y issue: ListSubHeader uses a `<p>` component inside a `<ul>`
- [ ] ListItem right icon is positioned using `position: absolute;`. When the screen is too small or we are using a inline container the icon will overlap the text (see screenshot below).

List #0

mjcarsjens:
**Closing issues**
- Closes #345
- Closes #402


components/List/src/ListItemText.tsx
```diff
@@ -29,15 +17,16 @@ export interface ListItemTextProps extends BaseProps {
...
+    'denhaag-list__item-text--with-secondary': secondary,
+  });
+
   return (
-    <MaterialListItemText {...props} primary={primary} secondary={secondary} disableTypography={true} classes={classes}>
-      {props.children}
-    </MaterialListItemText>
+    <div {...props} className={rootClassNames}>
+      {primary && <span className="denhaag-list__item-text-primary">{primary}</span>}
+      {secondary && <span className="denhaag-list__item-text-secondary">{secondary}</span>}
```
bddjong, line 28: The secondary item text is not styled according to figma. (Figma has all uppercase and grey-3)

components/List/src/List.tsx
```diff
@@ -1,33 +1,28 @@
...
+    'denhaag-list__wrapper--with-subheader': subheader,
+  });
   return (
-    <MaterialList {...materialProps} classes={classes}>
-      {props.children}
-    </MaterialList>
+    <div {...props} className={rootClassNames}>
+      {subheader}
+      <ul className="denhaag-list">{props.children}</ul>
+    </div>
```
bddjong, line 25: I'm not sure how I feel about this BEM structure. There is an element selector (`denhaag-list__wrapper`) higher than the component block (`denhaag-list`). This makes it feel as thought the List header is not part of the component.

proprietary/Components/src/denhaag/list.tokens.json
```diff
@@ -31,43 +30,31 @@
...
-          "color": { "value": "{denhaag.color.blue.3}" }
-        }
+        "margin-inline-end": { "value": "24px" },
+        "color": { "value": "{denhaag.color.grey.4}" }
       },
       "item-text": {
         "primary": {
           "color": { "value": "{denhaag.color.grey.4}" },
+          "font-size": { "value": "1rem" },
           "font-weight": { "value": "400" },
```
bddjong, line 40: `denhaag.typography.weight.regular`

proprietary/Components/src/denhaag/list.tokens.json
```diff
@@ -31,43 +30,31 @@
...
         }
       },
       "subheader": {
-        "font-weight": {
-          "value": "700"
-        },
-        "color": {
-          "value": "{denhaag.color.grey.4}"
-        }
+        "font-weight": { "value": "700" },
```
bddjong, line 55: `denhaag.typography.weight.bold`

bddjong (changes requested):

components/List/src/List.tsx
```diff
@@ -1,33 +1,28 @@
...
 }
 
 /**
  * Primary UI component for user interaction
  */
-export const List: React.FC<ListProps> = (props: ListProps) => {
-  const materialProps = { subheader: props.subheader };
-
+export const List: React.FC<ListProps> = ({ subheader, ...props }: ListProps) => {
+  const rootClassNames = clsx('denhaag-list__wrapper', {
```
bddjong, line 18: This is overwriting the `props.className`, so users are unable to use that property now.

bddjong (changes requested):

PR #797, opened 2022-03-02
Feature/#337 form label

bddjong:
**Todo**
- [ ] Deprecate `FormControl` MaterialUI components
- [ ] Deprecate Switch
- [ ] Remove obsolete MUI tests 
- [x] Add `form-field`
- [x] Add `form-field-label`
- [x] Add `form-field-description`
- [x] Add `form-field-description--error`

**Closing issues**
- #337


mjcarsjens (changes requested):
Alright, so I promised I'd also add my comments here, so here it goes:

- For adding the `--error` modifier we do indeed need some JS, as styling with `:invalid` and a sibling selector is not really possible with the different setups necessary and as the `:has` CSS selector is not widely available yet this seems like the only way to go (for now).
- I personally think we should just export one `FormField` component which bundles label and description, instead of also having seperate `FormFieldLabel` and `FormFieldDescription` components. 
- Current setup for `FormFieldLabel` where simply `props.children` is rendered inside the html `label` component has a high change of invalid HTML (as `label` can only contain [Phrasing Content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content)

mjcarsjens:
@bddjong I made an attempt at the changes as we discussed. 

- While it is in fact nicer to have a single component been having some issues with the necessary use of `id`'s in form elements today :(
- It is possible to just supply the input field in the `props.children`, but if you do this our `FormField` component needs to get a `htmlFor` to forward to the `label` and a `id` prop which gets forwarded to the Description. To avoid this I've been trying some magic, but not a big fan of my solution here either tbh. The input is passed through the `input` prop. The solution used now uses an actual `React` element, which is then checked whether it has an `id` and `aria-describedby`, and if not these are randomly generated (not the nicest random generator, I know, just as a PoC). Then a new element is created with the same type and props as our `input` with the generated `id` and `aria-describedby` if these didn't exist yet, which is then actually rendered. 
- Another solution which is practically the same but maybe cleaner wrt code: instead of having a `input` prop, have 2 props: `inputComponent` which accepts a `ReactType` and `inputProps`. This way you would get `inputComponent={TextField} inputProps={{placeholder: "Placeholder", icon: <CalendarIcon />}}`. I'm not sure whether I prefer this over the current solution, or whether I simply prefer actually just having 3 components...

mjcarsjens:
tbf also not quite sure how this solution would work/hold-up with checkboxes / radio groups..

bddjong:
> * Another solution which is practically the same but maybe cleaner wrt code: instead of having a `input` prop, have 2 props: `inputComponent` which accepts a `ReactType` and `inputProps`. This way you would get `inputComponent={TextField} inputProps={{placeholder: "Placeholder", icon: <CalendarIcon />}}`. I'm not sure whether I prefer this over the current solution, or whether I simply prefer actually just having 3 components...

@mjcarsjens I do actually prefer having separate components. In my opinion it keeps the code simpler and a more accurate representation of the actual DOM tree, instead of abstracting everything into props. There could be use cases for being able to add different components in between the input and description for example. This approach would not allow that.

components/FormField/src/index.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { HTMLAttributes } from 'react';
+import './index.scss';
+import clsx from 'clsx';
+
+export interface FormFieldProps extends HTMLAttributes<HTMLDivElement> {
+  label: string;
+  input: React.ReactNode;
```
bddjong, line 7: I think this should be `React.ReactElement`? ReactNode is throwing errors in my IDE for not defining some of the properties you're accessing below

PR #806, opened 2022-03-04
FormProgress Component

### LINKED ISSUE #728: OpenForms Steps progress
## Design 
https://www.figma.com/proto/MMQRIw6742Vg7REBV4Gnmh/Schetsen---MijnDenHaag-en-formulieren-2022?page-id=3007%3A21907&node-id=3007%3A22580&viewport=241%2C48%2C0.59&scaling=min-zoom&starting-point-node-id=3007%3A22580&hide-ui=1

### With previous link
image

### Without previous link
image

mjcarsjens:
@bddjong a few questions about this solution, could you look at these once you have the time?

- Not sure about this label setup, feels quite annoying to have there `stepLabel` and `ofLabel` props. Any ideas on how to fix this nicely? I considered using perhaps some sort of format string a la python (e.g. `label="Stap {0} van {1}") in a label prop, or alternatively just provide one `label` prop with no formatting whatsoever and let the end-user deal with it, but curious whether you have any thoughts on this.
- the `htmlFor` and `id` on the `label` and `progress` are currently static, resulting in invalid HTML/duplicate id's when the component is used multiple times. 

**Closing issues**
- Closes #728

### DIFF
--- a/components/FormProgress/src/index.tsx
+++ b/components/FormProgress/src/index.tsx
@@ -6,10 +6,9 @@ import './index.scss';
 export interface FormProgressProps extends HTMLAttributes<HTMLDivElement> {
   value: number;
   max: number;
+  label: string;
   previousHref?: string;
   previousLabel?: string;
-  stepLabel?: string;
-  ofLabel?: string;
 }
 
 /**
@@ -18,10 +17,10 @@ export interface FormProgressProps extends HTMLAttributes<HTMLDivElement> {
 export const FormProgress: React.FC<FormProgressProps> = ({
   value,
   max,
+  label,
   previousHref = 'javascript:history.back()',
   previousLabel = 'Vorige stap',
-  stepLabel = 'Stap',
-  ofLabel = 'van',
+  id = 'denhaag-form-progress-bar',
   ...props
 }: FormProgressProps) => {
   return (
@@ -34,16 +33,11 @@ export const FormProgress: React.FC<FormProgressProps> = ({
             </Link>
           </div>
         )}
-        <label className="denhaag-form-progress__label" htmlFor="denhaag-form-progress-bar">
-          {stepLabel} {value} {ofLabel} {max}
+        <label className="denhaag-form-progress__label" htmlFor={id}>
+          {label}
         </label>
       </div>
-      <progress
-        id="denhaag-form-progress-bar"
-        className="denhaag-form-progress__progress"
-        max={max}
-        value={value}
-      ></progress>
+      <progress id={id} className="denhaag-form-progress__progress" max={max} value={value}></progress>
     </div>
   );
 };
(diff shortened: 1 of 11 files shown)


components/FormProgress/src/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <div className="denhaag-form-progress__header">
+        {value > 1 && (
+          <div className="denhaag-form-progress__previous">
+            <Link href={previousHref} icon={<ArrowLeftIcon />} iconAlign="start">
+              {previousLabel}
+            </Link>
+          </div>
+        )}
+        <label className="denhaag-form-progress__label" htmlFor="denhaag-form-progress-bar">
+          {stepLabel} {value} {ofLabel} {max}
```
bddjong, line 38: I agree that this doesn't feel very nice. I think the component should not be so rigid about the label structure, but rather accept a simple `label` prop which can be formatted as desired by the applications using the component. Maybe the step indicator [POC for OpenForms](https://github.com/nl-design-system/open-forms-sdk/pull/4/files#diff-7916963985f6b54523951aa9bdcdb855db5905bed40fc7775a485986aec0fa5dR30) can give some inspiration on how to model the API for this component.

components/FormProgress/src/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+        </label>
+      </div>
+      <progress
+        id="denhaag-form-progress-bar"
```
bddjong, line 42: optional property for the `id` and `htmlFor` properties so they can be overridden if needed?

components/FormProgress/src/index.tsx
mjcarsjens: Not sure about the PR you just linked tbh, I literally used the exact same setup before with `currentStep` and `totalSteps` before defaulting to `value` and `max` as they are the HTML default attributes on a `progress` element. Only difference in that PR besides this naming is that they don't have i18n support by simply always printing `Stap {currentsStep} van {totalSteps}` right?

mjcarsjens: But yea, maybe still prefer the `currentStep` `totalSteps` naming, felt a bit clashing with our "try to follow HTML attribute naming when applicable"

components/FormProgress/src/index.tsx
mjcarsjens: Fair

components/FormProgress/src/index.tsx
bddjong: I linked it more as a context of the application we'll be using it in. So the component will be placed there instead of the hardcoded solution it has now.

mjcarsjens: Ahh like that, fair enough, thanks for the explanation!

mjcarsjens: Alrighty, changed it up, not sure whether I like this `label` either though, feels very inefficient to not have the `value` and `max` variables used somehow in the `label` string but this does support the most use-cases though

components/FormProgress/src/index.tsx
mjcarsjens: Fixed

bddjong (approved):

Issue #816, opened 2022-03-08
Anchor navigation


robbieatacato:
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1628%3A287

- html
- css

closes #816

PR #818, opened 2022-03-09
feat(anchor-navigation): add html, css, tokens

robbieatacato:
Solve: https://github.com/nl-design-system/denhaag/issues/816

Purpose:
- add html, css, tokens

closes #818

### DIFF
--- a/components/AnchorNavigation/package.json
+++ b/components/AnchorNavigation/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "@gemeente-denhaag/anchornavigation",
+  "version": "0.1.0",
+  "description": "An anchor navigation component",
+  "bugs": "https://github.com/nl-design-system/denhaag/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nl-design-system/denhaag.git",
+    "directory": "components/AnchorNavigation"
+  },
+  "license": "EUPL-1.2",
+  "author": "Municipality of The Hague",
+  "exports": "./dist/index.js",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c ../../rollup.config.js",
+    "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
+  }
+}
--- a/components/AnchorNavigation/package.json
+++ b/components/AnchorNavigation/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@gemeente-denhaag/anchornavigation",
+  "name": "@gemeente-denhaag/anchor-navigation",
   "version": "0.1.0",
   "description": "An anchor navigation component",
   "bugs": "https://github.com/nl-design-system/denhaag/issues",
--- a/components/AnchorNavigation/src/index.scss
+++ b/components/AnchorNavigation/src/index.scss
@@ -1,0 +1,46 @@
+.denhaag-anchor-navigation {
+  border-inline-start: var(--denhaag-anchor-navigation-border-inline-start);
+}
+
+.denhaag-anchor-navigation ul {
+  list-style: var(--denhaag-anchor-navigation-ul-list-style);
+  padding-inline-start: var(--denhaag-anchor-navigation-ul-padding-inline-start);
+}
+
+.denhaag-anchor-navigation ul li a {
+  font-family: var(--denhaag-anchor-navigation-anchor-font-family);
+  font-size: var(--denhaag-anchor-navigation-anchor-font-size);
+  line-height: var(--denhaag-anchor-navigation-anchor-line-height);
+  text-decoration: var(--denhaag-anchor-navigation-anchor-text-decoration);
+}
+
+.denhaag-anchor-navigation ul li a,
+.denhaag-anchor-navigation ul li a:visited {
+  color: var(--denhaag-anchor-navigation-anchor-color-default);
+}
+
+.denhaag-anchor-navigation ul li a:active,
+.denhaag-anchor-navigation ul li a:hover,
+.denhaag-anchor-navigation ul li a:target {
+  color: var(--denhaag-anchor-navigation-anchor-color-interactive);
+  position: relative;
+}
+
+.denhaag-anchor-navigation ul li a:active,
+.denhaag-anchor-navigation ul li a:target {
+  font-weight: var(--denhaag-anchor-navigation-anchor-font-weight);
+}
+
+.denhaag-anchor-navigation ul li a:active::before,
+.denhaag-anchor-navigation ul li a:hover::before,
+.denhaag-anchor-navigation ul li a:target::before {
+  content: "";
+  width: var(--denhaag-anchor-navigation-anchor-width);
+  height: var(--denhaag-anchor-navigation-anchor-line-height);
+  position: absolute;
+  top: calc(-1 * (0.3023 * var(--denhaag-anchor-navigation-anchor-line-height)));
+  left: calc(
+    (-1 * var(--denhaag-anchor-navigation-ul-padding-inline-start)) - var(--denhaag-anchor-navigation-anchor-width)
+  );
+  background-color: var(--denhaag-anchor-navigation-anchor-background-color);
+}
--- a/components/AnchorNavigation/src/index.scss
+++ b/components/AnchorNavigation/src/index.scss
@@ -2,49 +2,49 @@
   border-inline-start: var(--denhaag-anchor-navigation-border-inline-start);
 }
 
-.denhaag-anchor-navigation .denhaag-anchor-navigation__list {
+.denhaag-anchor-navigation__list {
   list-style: var(--denhaag-anchor-navigation-list-list-style);
   padding-inline-start: var(--denhaag-anchor-navigation-list-padding-inline-start);
 }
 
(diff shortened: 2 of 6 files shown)


components/AnchorNavigation/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@gemeente-denhaag/anchornavigation",
```
paulAcato, line 2: je gebruikt overal `anchor-navigation` wellicht hier ook toepassen?

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,46 @@
+.denhaag-anchor-navigation {
+  border-inline-start: var(--denhaag-anchor-navigation-border-inline-start);
+}
+
+.denhaag-anchor-navigation ul {
```
paulAcato, line 5: Apply as BEM CSS classes ;-) for example: `.denhaag-anchor-navigation__list`

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,46 @@
...
+  list-style: var(--denhaag-anchor-navigation-ul-list-style);
+  padding-inline-start: var(--denhaag-anchor-navigation-ul-padding-inline-start);
+}
+
+.denhaag-anchor-navigation ul li a {
```
paulAcato, line 10: This kind of nesting is unnecassary,  `.denhaag-anchor-navigation ul li a` should do the same as `.denhaag-anchor-navigation a`

Or if u never want any issue with it apply a CSS selector: `.denhaag-anchor-navigation__link`

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,46 @@
...
+  font-family: var(--denhaag-anchor-navigation-anchor-font-family);
+  font-size: var(--denhaag-anchor-navigation-anchor-font-size);
+  line-height: var(--denhaag-anchor-navigation-anchor-line-height);
+  text-decoration: var(--denhaag-anchor-navigation-anchor-text-decoration);
+}
+
+.denhaag-anchor-navigation ul li a,
+.denhaag-anchor-navigation ul li a:visited {
```
paulAcato, line 18: No modifier classes for storybook? My suggestion: `.denhaag-anchor-navigation__link--visited`

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,46 @@
...
+  color: var(--denhaag-anchor-navigation-anchor-color-default);
+}
+
+.denhaag-anchor-navigation ul li a:active,
```
paulAcato, line 22: See comment above

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,46 @@
...
+.denhaag-anchor-navigation ul li a:hover,
+.denhaag-anchor-navigation ul li a:target {
+  color: var(--denhaag-anchor-navigation-anchor-color-interactive);
+  position: relative;
+}
+
+.denhaag-anchor-navigation ul li a:active,
```
paulAcato, line 29: See comment above

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,46 @@
...
+.denhaag-anchor-navigation ul li a:target {
+  font-weight: var(--denhaag-anchor-navigation-anchor-font-weight);
+}
+
+.denhaag-anchor-navigation ul li a:active::before,
```
paulAcato, line 34: Why not always add the `:before` and just change the background color on one of those CSS selectors.

components/AnchorNavigation/src/stories/bem.stories.mdx
```diff
@@ -0,0 +1,71 @@
...
+## Stories
+
+### Default
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="Default">
+    <nav class="denhaag-anchor-navigation" aria-labelledby="secondary-navigation">
+      <ul id="secondary-navigation">
+        <li>
```
paulAcato, line 33: I'm missing all the CSS selectors :) This way there are no modifiers possible.

proprietary/Components/src/denhaag/anchor-navigation.tokens.json
```diff
@@ -0,0 +1,22 @@
...
+  "denhaag": {
+    "anchor-navigation": {
+      "border-inline-start": { "value": "2px solid {denhaag.color.grey.2}" },
+      "ul": {
+        "list-style": { "value": "none" },
+        "padding-inline-start": { "value": "1.125rem" }
+      },
+      "anchor": {
+        "background-color": { "value": "{denhaag.color.green.3}" },
+        "color-default": { "value": "{denhaag.color.grey.4}" },
```
paulAcato, line 11: 1: `color` should be enough instead of `color-default`
2: By defining the `color-interactive` this way you make a design choice, therefor i suggest a hover, focus & visited state in the tokens.

paulAcato (changes requested):

components/AnchorNavigation/src/index.scss
robbieatacato: This one is unnecessary in my opinion, because it's a browser thing when link is clicked, no need for using js to add this class.
Why add an exra css class if a link is used/clicked/visited?

components/AnchorNavigation/src/index.scss
robbieatacato: Because the grey border is set on the main component itself, the nav element for design reasons.
The border for the links are not property of the nav design.
These are only relevant when interacting with the links.

components/AnchorNavigation/src/index.scss
bddjong: > This one is unnecessary in my opinion, because it's a browser thing when link is clicked, no need for using js to add this class. Why add an exra css class if a link is used/clicked/visited and handled by the browser style?

Even though you're writing tests at the moment, it's easier to do visual regression tests when there is a class name for the visited effect. We can write a test using `--visited` to ensure the pseudo selectors don't visually change.

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,57 @@
+.denhaag-anchor-navigation {
+  border-inline-start: var(--denhaag-anchor-navigation-border-inline-start);
+}
+
+.denhaag-anchor-navigation .denhaag-anchor-navigation__list {
```
paulAcato, line 5: Just `.denhaag-anchor-navigation__list {` should be enough

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,57 @@
...
+  list-style: var(--denhaag-anchor-navigation-list-list-style);
+  padding-inline-start: var(--denhaag-anchor-navigation-list-padding-inline-start);
+}
+
+.denhaag-anchor-navigation .denhaag-anchor-navigation__link {
```
paulAcato, line 10: Just `.denhaag-anchor-navigation__link {` should be enough

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,57 @@
...
+  font-family: var(--denhaag-anchor-navigation-link-font-family);
+  font-size: var(--denhaag-anchor-navigation-link-font-size);
+  line-height: var(--denhaag-anchor-navigation-link-line-height);
+  text-decoration: var(--denhaag-anchor-navigation-link-text-decoration);
+}
+
+.denhaag-anchor-navigation .denhaag-anchor-navigation__link,
```
paulAcato, line 17: `.denhaag-anchor-navigation__link,
.denhaag-anchor-navigation__link:visited {` should be enough

components/AnchorNavigation/src/index.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.denhaag-anchor-navigation .denhaag-anchor-navigation__link:hover {
+  color: var(--denhaag-anchor-navigation-link-color-hover);
+}
+
+.denhaag-anchor-navigation .denhaag-anchor-navigation__link:target {
+  color: var(--denhaag-anchor-navigation-link-color-target);
+}
+
```
paulAcato, line 33: Since it's one property it isn't bad but just a suggestion since we're working with CSS variables:


`.denhaag-anchor-navigation__link:target { --denhaag-anchor-navigation-link-color:  var(--denhaag-anchor-navigation-link-color-target) }`

paulAcato (changes requested):

components/AnchorNavigation/src/stories/bem.stories.mdx
robbieatacato: There is no reason to style the li's

components/AnchorNavigation/src/index.scss
robbieatacato: @bddjong @paulAcato  also when the styling is the same for the default anchor link and a:visited?
Otherwise, I wil add additional BEM html with no difference with the default.

robbieatacato:
@bddjong I made changes with additional bem html for the .denhaag-anchor-navigation__link--visited class

components/AnchorNavigation/src/stories/bem.stories.mdx
paulAcato: That was an old commend before u added className at all ;-)

paulAcato (approved):

Issue #837, opened 2022-03-21
Link group


robbieatacato:
- Figma: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1485%3A1156

closes #837

PR #839, opened 2022-03-21
feat(link-group): add html, css, tokens

robbieatacato:
Solve: https://github.com/nl-design-system/denhaag/issues/837

Purpose
Adding Link Group:
- html
- css
- tokens

closes #839


components/LinkGroup/src/stories/bem.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+### Default
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="Default">
+    <div class="denhaag-link-group">
+      <h3 class="utrecht-heading-4 denhaag-link-group__caption">Caption</h3>
+      <nav class="denhaag-link-group__navigation" aria-labelledby="denhaag-link-group-navigation-1">
+        <ul
+          id="denhaag-link-group-navigation-1"
```
paulAcato, line 35: Not sure if this needs an ID?

packages/components-css/src/index.tsx
```diff
@@ -1,12 +1,10 @@
 import '../../../components/Alert/src/index.scss';
 import '../../../components/Accordion/src/index.scss';
 import '../../../components/AnchorNavigation/src/index.scss';
-import '../../../components/AuthenticationBlock/src/index.scss';
```
paulAcato, line 4: Why are you removing blocks which area already on the `main` branch?

packages/components-css/src/index.tsx
```diff
@@ -1,12 +1,10 @@
...
 import '../../../components/BadgeCounter/src/index.scss';
 import '../../../components/Blockquote/src/index.scss';
 import '../../../components/Breadcrumb/src/index.scss';
 import '../../../components/Button/src/index.scss';
-import '../../../components/ButtonGroup/src/index.scss';
```
paulAcato, line 9: Why are you removing blocks which area already on the `main` branch?

paulAcato (changes requested):

components/LinkGroup/src/stories/bem.stories.mdx
robbieatacato: @paulAcato Its for screen readers, see also MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#labeling_section_content

packages/components-css/src/index.tsx
robbieatacato: fixt it

components/LinkGroup/src/stories/bem.stories.mdx
paulAcato: @robbieatacato check, looked like a bit of overkill to me ;-)

paulAcato (approved):

components/LinkGroup/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@gemeente-denhaag/link-group",
+  "description": "The Link Group component. List of links.",
+  "author": "Municipality of The Hague",
+  "license": "EUPL-1.2",
+  "version": "0.1.1-alpha.46",
```
bddjong, line 6: We shouldn't start versioning at 46 for new components.

proprietary/Components/src/denhaag/link-group.tokens.json
```diff
@@ -0,0 +1,20 @@
...
+    "link-group": {
+      "image": {
+        "padding-block-end": { "value": "1rem" }
+      },
+      "caption": {
+        "color": { "value": "{denhaag.color.grey.5}" },
+        "font-weight": { "value": "500" }
+      },
+      "navigation-link": {
+        "line-height": { "value": "1.6875rem" }
```
bddjong, line 12: If the link is 18px, and line height is 27px, we should use `"value": "1.5"` in my opinion.

components/LinkGroup/README.md
```diff
@@ -0,0 +1,154 @@
+# Link Group
```
bddjong, line 1: This documentation is titled link group, but a quick read tells me it's only talking about the Link itself, and nothing about link groups.

bddjong (commented):
Still have some comments. I don't think there's anything we can do about the version number now though, since it's already merged.

PR #841, opened 2022-03-22
Feature/gdh 200 breadcrumbs feedback

paulAcato:
Applied max-width and background to the breadcrumbs. Apparenly one of my screens doesn't show this color, therefor I completely missed it.

### DIFF
--- a/components/Breadcrumb/src/index.scss
+++ b/components/Breadcrumb/src/index.scss
@@ -1,18 +1,32 @@
 .denhaag-breadcrumb {
-  --denhaag-breadcrumb-background-color: transparent;
+  --denhaag-breadcrumb-link-background-color: transparent;
 
+  background-color: var(--denhaag-breadcrumb-background-color, transparent);
   color: var(--denhaag-breadcrumb-color, inherit);
+  padding-block-end: var(--denhaag-breadcrumb-padding-block);
+  padding-block-start: var(--denhaag-breadcrumb-padding-block);
+  padding-inline-end: var(--denhaag-breadcrumb-padding-inline);
+  padding-inline-start: var(--denhaag-breadcrumb-padding-inline);
+}
+
+@media (min-width: 1124px) {
+  .denhaag-breadcrumb {
+    --denhaag-breadcrumb-padding-inline: 0;
+  }
 }
 
 .denhaag-breadcrumb__list {
   block-size: 100%;
   display: flex;
+  margin-block-end: 0;
+  margin-block-start: 0;
+  margin-inline-end: auto;
+  margin-inline-start: auto;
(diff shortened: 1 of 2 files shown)


components/Breadcrumb/src/index.scss
```diff
@@ -1,18 +1,32 @@
...
 }
 
 .denhaag-breadcrumb__list {
   block-size: 100%;
   display: flex;
+  margin-block-end: 0;
+  margin-block-start: 0;
+  margin-inline-end: auto;
+  margin-inline-start: auto;
+  max-width: var(--denhaag-breadcrumb-max-width, 100%);
```
robbieatacato, line 25: If this `--denhaag-breadcrumb-max-width` is only used for the `__list` modifier, than the token can be named `--denhaag-breadcrumb-list-max-width`

components/Breadcrumb/src/index.scss
```diff
@@ -23,11 +37,11 @@ ol.denhaag-breadcrumb__list {
 }
 
 .denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child) {
-  --denhaag-breadcrumb-background-color: var(--denhaag-breadcrumb-dots-background-color, inherit);
+  --denhaag-breadcrumb-link-background-color: var(--denhaag-breadcrumb-dots-background-color, inherit);
```
robbieatacato, line 40: `inherit` is not necessary, because the default fallback is `inherit` when the value could not be read.

components/Breadcrumb/src/index.scss
```diff
@@ -23,11 +37,11 @@ ol.denhaag-breadcrumb__list {
...
 }
 
 .denhaag-breadcrumb__item:not(:first-child):not(:nth-last-child(2)):not(:last-child):hover {
-  --denhaag-breadcrumb-background-color: var(--denhaag-breadcrumb-dots-hover-background-color, inherit);
+  --denhaag-breadcrumb-link-background-color: var(--denhaag-breadcrumb-dots-hover-background-color, inherit);
```
robbieatacato, line 44: idem dito

robbieatacato (changes requested):

components/Breadcrumb/src/index.scss
paulAcato: This logic has been approved before, currently only the name has been changed. 

Therefor I'll not change the logic. Overwritting an existing CSS variable with a new one requires a back-up variable. Otherwise it's possible to be undefined, therefor the whole variable we cast to is also undefined.

components/Breadcrumb/src/index.scss
paulAcato: Edited to `--denhaag-breadcrumb-list-max-width`

Issue #876, opened 2022-04-04
Hero - theme page


robbieatacato:
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=11810%3A67677

PR #878, opened 2022-04-05
feat(hero-theme-page): add hero variant, theme page

robbieatacato:
Solve: https://github.com/nl-design-system/denhaag/issues/876

Purpose:
- Add new variant, theme page

closes #878

### DIFF
--- a/components/Hero/src/index.scss
+++ b/components/Hero/src/index.scss
@@ -30,6 +30,15 @@
   overflow: var(--denhaag-hero-theme-page-overflow);
 }
 
+.denhaag-hero__image-pattern {
+  position: absolute;
+}
+
+.denhaag-hero--theme-page .denhaag-hero__image-pattern {
+  height: 100%;
+  width: auto;
+}
+
 .denhaag-hero__container {
   display: grid;
   grid-auto-flow: dense;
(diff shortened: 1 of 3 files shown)


components/Hero/src/index.scss
```diff
@@ -8,7 +8,8 @@
   position: relative;
 }
 
-.denhaag-hero--routing {
+.denhaag-hero--routing,
+.denhaag-hero--theme-page {
```
paulAcato, line 12: Personally i'm more a fan of using `--theme` instead of `--theme-page` since a hero is kinda always used on a page ;-) Otherwise the other hero's must be modified with `-page` behind it.

@bddjong what are your thoughts?

proprietary/Components/src/denhaag/hero.tokens.json
```diff
@@ -171,6 +171,41 @@
             "value": "16rem"
           }
         }
+      },
+      "theme-page": {
+        "height": { "value": "288px" },
```
paulAcato, line 176: In NLDS TheHague they use rems instead of pixels, please continue to to use rems :)

paulAcato (changes requested):

components/Hero/src/index.scss
bddjong: I think this is actually a mistranslation regardless. I had to run the storybook to understand what kind of variant we were actually talking about. For "thema paginas", thema refers to a topic or subject rather than a theme. Especially in the context of css, calling them `theme` is very confusing.

bddjong (changes requested):

components/Hero/src/index.scss
paulAcato: The naming `theme` comes from Figma. If `theme` isn't correct or `theme-page` shouldn't we need to discuss this with design?

bddjong: Also a good point. I agree we should discuss this with design to have consistent naming for this both in different components and between figma and storybook.

paulAcato (approved):

PR #882, opened 2022-04-07
style(hero-theme-page): refactor bem, css + tokens

robbieatacato:
Solve: https://github.com/nl-design-system/denhaag/issues/876

Purpose:
- Refactor bem, css, tokens more to design likings.

Screen capture:

https://user-images.githubusercontent.com/95216123/162232935-44739949-c568-44c5-ae3b-3db30c4be1ea.mov

closes  #882


components/Hero/src/index.scss
```diff
@@ -171,6 +192,43 @@
...
+  }
+}
+
+@media (min-width: 430px) {
+  .denhaag-hero--theme-page {
+    height: var(--denhaag-hero-theme-page-height-s);
+  }
+
+  .denhaag-hero--theme-page .denhaag-hero__text {
+    font-size: var(--denhaag-hero-theme-page-text-font-size-m);
```
paulAcato, line 220: Why not do the same as the other hero's and use the power of CSS variables?

For example:

`@media (min-width: 430px) { 
 .denhaag-hero--theme-page { 
   --denhaag-hero-theme-page-height: var(--denhaag-hero-theme-page-height-s);
   --denhaag-hero-theme-page-text-font-size:  var(--denhaag-hero-theme-page-text-font-size-m); 
 }
}
`

This way u shorten the CSS output and use the way CSS variables are meant to be,

components/Hero/src/index.scss
```diff
@@ -171,6 +192,43 @@
...
     --denhaag-hero-content-column-end: var(--denhaag-hero-home-content-column-end-xs, 12);
   }
+
+  .denhaag-hero--theme-page .denhaag-hero__container {
+    padding-block-start: var(--denhaag-hero-theme-page-container-padding-block-start-s);
+  }
+
+  .denhaag-hero--theme-page .denhaag-hero__content {
+    min-height: var(--denhaag-hero-theme-page-content-min-height-s);
+    height: var(--denhaag-hero-theme-page-content-height-s);
```
paulAcato, line 202: Why a height and min-height? 

When a height is set, a min-height is redundant.

proprietary/Components/src/denhaag/hero.tokens.json
```diff
@@ -342,36 +342,65 @@
...
         "title": {
-          "padding-block-end": { "value": "{denhaag.space.block.md}" }
+          "padding-block-end": { "value": "{denhaag.space.block.md}" },
+          "font-size-s": { "value": "8.9vw" },
+          "font-size-m": { "value": "3rem" }
         },
         "text": {
           "font-size": { "value": "{denhaag.typography.scale.lg.font-size}" },
-          "font-size-s": { "value": "{denhaag.typography.scale.xl.font-size}" },
+          "font-size-s": { "value": "5.6vw" },
```
paulAcato, line 402: Please use the typography tokens, font-sizes shouldn't be set on viewport width.

And for the line below, im pretty sure there's a 1.5rem typography token.

paulAcato (changes requested):
I'm missing the container being aligned on the grid. Like we did with the other hero's. 

The grid-alignment with grid-start and grid-end is perfectly suited for this.

components/Hero/src/index.scss
robbieatacato: That is necessary because the height is responsive, and these have different values.

proprietary/Components/src/denhaag/hero.tokens.json
robbieatacato: The typography tokens don't have vw's voor responsive font-sizes.

paulAcato: I checked this with Joni last tuesday. We don't use VW properties. The font-sizes should always be from the tokens conform the defined properties. If Figma says to use VW, please contact the design team.

paulAcato (changes requested):
See comment about font-size vw

paulAcato (changes requested):

paulAcato (approved):

PR #889, opened 2022-04-12
Feature/gdh 261 typography spacing

paulAcato:
Added the formula for headings when after another element.


paulAcato:
@bddjong, i notice there's also work in this PR which was also in the former PR about typography (which was approved and  merged to main (some days ago))?

Issue #903, opened 2022-04-19
Highlighted links


robbieatacato:
Solve: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=0%3A1

PR #906, opened 2022-04-20
feat(highlighted-links): add bem, css, tokens

robbieatacato:
Solve: https://github.com/nl-design-system/denhaag/issues/903

Purpose:
- Highlighted links are important links at the top of your page.
- add bem, css, tokens for highlighted links

closes #906

### DIFF
--- a/components/HighlightedLinks/src/index.scss
+++ b/components/HighlightedLinks/src/index.scss
@@ -1,0 +1,41 @@
+.denhaag-highlighted-links {
+  background-color: var(--denhaag-highlighted-links-background-color);
+  padding-block-start: var(--denhaag-highlighted-links-padding-block-start);
+  padding-block-end: var(--denhaag-highlighted-links-padding-block-end);
+  width: var(--denhaag-highlighted-links-width);
+}
+
+.denhaag-highlighted-links__container {
+  --denhaag-highlighted-links-width: var(--denhaag-highlighted-links-container-width);
+
+  margin-inline-start: var(--denhaag-highlighted-links-container-margin-inline-start);
+  margin-inline-end: var(--denhaag-highlighted-links-container-margin-inline-end);
+  max-width: var(--denhaag-highlighted-links-container-max-width);
+}
+
+.denhaag-hero--theme-page + .denhaag-highlighted-links {
+  --denhaag-highlighted-links-padding-block-start: var(
+    --denhaag-highlighted-links-hero-theme-page-highlighted-links-padding-block-start
+  );
+}
+
+.denhaag-highlighted-links__navigation-list {
+  display: var(--denhaag-highlighted-links-navigation-list-display);
+  gap: var(--denhaag-highlighted-links-navigation-list-gap);
+
+  @media (min-width: 48rem) {
+    grid-template-columns: var(--denhaag-highlighted-links-navigation-list-grid-template-columns);
+  }
+
+  @media (min-width: 68.875rem) {
+    --denhaag-highlighted-links-navigation-list-grid-template-columns: var(
+      --denhaag-highlighted-links-navigation-list-grid-template-columns-2
+    );
+  }
+
+  @media (min-width: 80rem) {
+    --denhaag-highlighted-links-navigation-list-grid-template-columns: var(
+      --denhaag-highlighted-links-navigation-list-grid-template-columns-3
+    );
+  }
+}
--- a/components/HighlightedLinks/src/index.scss
+++ b/components/HighlightedLinks/src/index.scss
@@ -8,18 +8,24 @@
 .denhaag-highlighted-links__navigation-list {
   display: var(--denhaag-highlighted-links-navigation-list-display);
   gap: var(--denhaag-highlighted-links-navigation-list-gap);
+}
 
-  @media (min-width: 48rem) {
+@media (min-width: 48rem) {
+  .denhaag-highlighted-links__navigation-list {
     grid-template-columns: var(--denhaag-highlighted-links-navigation-list-grid-template-columns);
   }
+}
 
-  @media (min-width: 68.875rem) {
+@media (min-width: 68.875rem) {
+  .denhaag-highlighted-links__navigation-list {
     --denhaag-highlighted-links-navigation-list-grid-template-columns: var(
       --denhaag-highlighted-links-navigation-list-grid-template-columns-2
     );
   }
+}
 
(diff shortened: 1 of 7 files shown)


proprietary/Components/src/denhaag/highlighted-links.tokens.json
```diff
@@ -0,0 +1,26 @@
...
+    "highlighted-links": {
+      "background-color": { "value": "{denhaag.color.warmgrey.1}" },
+      "padding-block-start": { "value": "{denhaag.space.block.3xl}" },
+      "padding-block-end": { "value": "{denhaag.space.block.3xl}" },
+      "width": { "value": "100%" },
+      "container": {
+        "margin-inline-start": { "value": "auto" },
+        "margin-inline-end": { "value": "auto" },
+        "width": { "value": "100%" },
+        "max-width": { "value": "98rem" }
```
paulAcato, line 12: If this component is within a parent component like the page-container, this property redundant. Will this component used outside or inside a parent component? Otherwise there will be 2 containers colliding.

components/HighlightedLinks/src/index.scss
```diff
@@ -0,0 +1,41 @@
...
+
+.denhaag-highlighted-links__navigation-list {
+  display: var(--denhaag-highlighted-links-navigation-list-display);
+  gap: var(--denhaag-highlighted-links-navigation-list-gap);
+
+  @media (min-width: 48rem) {
+    grid-template-columns: var(--denhaag-highlighted-links-navigation-list-grid-template-columns);
+  }
+
+  @media (min-width: 68.875rem) {
```
paulAcato, line 30: Overwrites of a CSS var u want to do on the top level of the element, that way the CSS output is reduced and other developers can overwrite these values more easily. #truepowerofcssvariables.

proprietary/Components/src/denhaag/highlighted-links.tokens.json
```diff
@@ -0,0 +1,26 @@
...
+      },
+      "hero-theme-page-highlighted-links": {
+        "padding-block-start": { "value": "4.375rem" }
+      },
+      "navigation-list": {
+        "display": { "value": "grid" },
+        "gap": { "value": "0 {denhaag.space.block.3xl}" },
+        "grid-template-columns": { "value": "repeat(auto-fill, minmax(22.75rem, 1fr))" },
```
paulAcato, line 20: Correct me if i'm wrong. But if i check the design it's always 1, 2 or 3 columns. if thats the case, it's unnecessary to use a minmax()

components/HighlightedLinks/src/index.scss
```diff
@@ -0,0 +1,41 @@
+.denhaag-highlighted-links {
+  background-color: var(--denhaag-highlighted-links-background-color);
+  padding-block-start: var(--denhaag-highlighted-links-padding-block-start);
+  padding-block-end: var(--denhaag-highlighted-links-padding-block-end);
+  width: var(--denhaag-highlighted-links-width);
+}
+
+.denhaag-highlighted-links__container {
```
paulAcato, line 8: This block will be edited from the editor, therefor it's already within a container. I believe this container is redundant. Please check with design.

paulAcato (changes requested):
Please check with design and where we're using this block.

components/HighlightedLinks/src/index.scss
```diff
@@ -0,0 +1,41 @@
+.denhaag-highlighted-links {
```
paulAcato, line 1: This component must be edited from wordpress content editor. Therefor it's already within a container and must breakout of the page container, due the background is fullwidth.

My suggestion would be to make this element position:relative, and set the background-color on the after u make 100vw. 

This way the container is always the same width as the parent-container.

paulAcato (changes requested):

proprietary/Components/src/denhaag/highlighted-links.tokens.json
robbieatacato: @paulAcato UX had specific column amount in mind per breakpoint.
There for scaling of column minimum widths is necessary, otherwise it won't behave with the specific amount of columns per breakpoint.

paulAcato: `grid-template-columns: repeat(3, 1fr);` no minmax required for this ;-)

robbieatacato: I tested it, does'nt work without it.

robbieatacato: I tested it without the minmax, does not work with specific breakpoints.

components/HighlightedLinks/src/index.scss
robbieatacato: @paulAcato This does not work in storybook:
```
  @media (min-width: var(--denhaag-highlighted-links-navigation-list-min-width)) {
    grid-template-columns: var(--denhaag-highlighted-links-navigation-list-grid-template-columns);
  }

  @media (--denhaag-highlighted-links-navigation-list-min-width: --denhaag-highlighted-links-navigation-list-min-width-2) {
    --denhaag-highlighted-links-navigation-list-grid-template-columns: var(
      --denhaag-highlighted-links-navigation-list-grid-template-columns-2
    );
  }
```

components/HighlightedLinks/src/index.scss
robbieatacato: I removed the container, the tokens and the css related to it.

components/HighlightedLinks/src/index.scss
paulAcato: This isn't supported on the web. Also thats not what i ment with my comment above.

What you want is the media-queries on the top level of the element ( `.denhaag-highlighted-links` ) instead of `.denhaag-highlighted-links__navigation-list`. The reason why is because if another develop want to use this component they need to overwrite it with `.denhaag-highlighted-links .denhaag-highlighted-links__navigation-list` before the css variable will react.

proprietary/Components/src/denhaag/highlighted-links.tokens.json
paulAcato: Check how i did it within the theme ;)

paulAcato (changes requested):

bddjong (approved):

PR #907, opened 2022-04-21
build: generate `dist/index.css` for components

Robbert:
In addition to the CSS that is bundled with the React component and automatically injected in the page, we can provide the CSS files for projects that don't use the React framework. This draft PR contains a proposal for how to add `dist/index.css` to the npm package. When we agree on an approach, I can update the PR to update the build script for all components.

### DIFF
--- a/components/ProcessSteps/package.json
+++ b/components/ProcessSteps/package.json
@@ -18,7 +18,9 @@
     "dist"
   ],
   "scripts": {
-    "build": "rollup -c ../../rollup.config.js",
+    "build": "npm-run-all build:**",
+    "build:css": "sass --no-source-map src/index.scss:dist/index.css",
+    "build:js": "rollup -c ../../rollup.config.js",
     "clean": "yarn rimraf dist tsconfig.tsbuildinfo"
   },
   "dependencies": {


Robbert:
I should add that the alternative approach I tried was to add `extract: true` to the `postcss()` config in `rollup.config.js`. This also works to expose `dist/index.css`, but then the React component would no longer include the automatically injected CSS: that would be a significant and breaking change.

React injecting inline CSS might cause issues when you do not allow `style-src: 'unsafe-inline'`, for example when working with DigiD security requirements. I'm not entirely sure about this. But it could be that in the future there is a special use case for React components without bundled CSS, and that we either switch to that or provide an additional build output that has `extract: true`.

mjcarsjens:
@Robbert the issue of inline styles and CSP headers is indeed something that is still on our radar, we should spend some time on investigating this further. I looked into the `extra: true` option for `postcss` in the `rollup.config.js` too a while ago but I didn't have a lot of time for this so when it wasn't working we postponed this for a bit.

For now at least providing the built CSS file in our packages seems like a nice addition to the library, but again, not a reason for us to not look into the inline-styles issue with react.

mjcarsjens (approved):

PR #908, opened 2022-04-21
style(hero-theme-page): height scaling hero theme-page, depending on …

robbieatacato:
Solve: The height of the white block depends on the size of the content.

Purpose: scalability.

closes #908

https://user-images.githubusercontent.com/95216123/164455859-8e77ba98-aa0a-4730-8a73-1538ba5390e5.mov

https://user-images.githubusercontent.com/95216123/164455861-4090eae4-0a72-4763-b31d-3a89e7d82853.mov


components/Hero/src/index.scss
```diff
@@ -198,7 +190,8 @@
     );
   }
 
-  .denhaag-hero--home {
+  .denhaag-hero--home,
```
paulAcato, line 193: Gescheiden houden!

paulAcato (changes requested):
Keep modifiers separate

components/Hero/src/index.scss
paulAcato: What you did below was applying a `-home` token into `-theme` and `-home`.
Everything that is specifically within the modifier of that hero must be within their own selector.
if the token for `-home` changes you don't want the `theme` to be updated. 
Im defintly not a fan of @extended. One of the reasons is pure readability in the code.  The @extended is used at the link-component. If u check the build-css it's defined 3 times. Therefor, each own modifier has their own tokens.

paulAcato (changes requested):

PR #918, opened 2022-04-25
build: provide CommonJS exports in addition to ES Modules

Robbert:
Experimenting with a method to provide the React components using CommonJS exports too.


bddjong:
@Robbert do you know whether this would break existing imports of components in the `nl-portal-libraries` for example? Or is the correct mjs/cjs index picked by all common bundlers. 

And in what instance would CommonJS be required? I thought ESM was the official standard.

Robbert:
> @Robbert do you know whether this would break existing imports of components in the `nl-portal-libraries` for example?

@bddjong I'm was trying to test a new local version of `@gemeente-denhaag/components-react` in `nl-portal-libraries`, but the dependency on the component library there is _ancient_: https://github.com/Gemeente-DenHaag/nl-portal-libraries/blob/master/packages/user-interface/package.json#L51 Switching to local is not easily possible because of so many breaking changes.

So while I couldn't test that, I did update switch `@gemeente-denhaag/process-steps` to a local build and that works fine. The `user-interface` build also generates CommonJS and ESM versions:

`index.js`:

```js
import{StepList as ke,Step as ye,StepSection as we,StepHeader as Me,StepMarker as Le,StepHeading as Te,StepExpandedIcon as Ce,SubStepList as Pe,SubStep as Ee,SubStepMarker as je,SubStepHeading as Ze}from"@gemeente-denhaag/process-steps";
```

and `index.modern.js`:

```js
p=require("@gemeente-denhaag/process-steps")
```

So the which version actually of the component (cjs or mjs) to run the app is dependent on the runtime. The only thing `yarn build` really needs is `index.d.ts` and that functionality also still works.

So I think we can safely update. But if you'd feel more confident taking a gradual approach, we could start with just this process steps component and update `nl-portal-libraries`. When it works, then we move forward with all the other packages.

---

I did run into a couple of issues starting the `nl-portal-libraries` project. Installing the local depenency only worked using `npx yarn@1.19.0 add` and not with `yarn add` with version 1.22.18:

```
error An unexpected error occurred: "expected workspace package to exist for \"jest-snapshot\"".
```

Also, the component library now appears to export multiple `Step` components, caused by:

```js
export * from '@gemeente-denhaag/process-steps';
// ...
export * from '@gemeente-denhaag/timeline';
```

bddjong:
@Robbert Thanks for testing that and the extra info. That should help us make a decision on how to support this. For me, it sounds OK to support both exports for all components. I'm guessing we'll first need to update the `nl-portal-libraries` to the latest version so we can properly test the impact of providing both exports for all components.

@mjcarsjens what's your opinion on this?

bddjong:
I've created issue #924 for the duplicate step components

mjcarsjens:
@bddjong @Robbert looks good to me, seems like a nice extra to also support CommonJS imports. Seems to me that when the import does work for the pds component there should be no issues for the other packages in our library, but would be nice if we could test this a bit further (and besides, even without these changes, the nl-portal-libraries project really should update its dependency to our component library a bit more often)

PR #928, opened 2022-04-26
feat(link-button): accessibility sr-only to button and link

paulAcato:
Added 2 accessibility points:

1. Buttons and links to specific cursors (advised by Fernand);
2. Add screen-reader inner-element support.

### DIFF
--- a/components/Button/src/index.scss
+++ b/components/Button/src/index.scss
@@ -4,7 +4,7 @@
   border: 0;
   border-radius: var(--denhaag-button-border-radius);
   color: var(--denhaag-button-primary-action-color);
-  cursor: pointer;
+  cursor: default;
   display: inline-flex;
   font-family: var(--denhaag-button-font-family);
   font-size: var(--denhaag-typography-scale-base-font-size);
@@ -141,4 +141,21 @@
   display: inherit;
   margin-inline-end: -4px;
   margin-inline-start: 8px;
+}
+
+.denhaag-button__sr-only {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding-block-start: 0;
+  padding-block-end: 0;
+  padding-inline-start: 0;
+  padding-inline-end: 0;
+  margin-block-start: -1px;
+  margin-block-end: -1px;
+  margin-inline-start: -1px;
+  margin-inline-end: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  border: 0;
 }
--- a/components/Button/src/index.scss
+++ b/components/Button/src/index.scss
@@ -4,7 +4,7 @@
   border: 0;
   border-radius: var(--denhaag-button-border-radius);
   color: var(--denhaag-button-primary-action-color);
-  cursor: default;
+  cursor: var(--denhaag-button-cursor, default);
   display: inline-flex;
   font-family: var(--denhaag-button-font-family);
   font-size: var(--denhaag-typography-scale-base-font-size);
@@ -15,6 +15,10 @@
   padding-inline-start: var(--denhaag-button-medium-size-padding-inline, var(--denhaag-button-padding-inline));
   position: relative;
   text-decoration: none;
+}
+
+a.denhaag-button {
+  --denhaag-button-cursor: pointer;
 }
 
 .denhaag-button--focus::after,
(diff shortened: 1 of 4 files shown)


components/Button/src/stories/bem.stories.mdx
```diff
@@ -126,3 +126,20 @@ import readme from "../../README.md";
   </Story>
   {/* eslint-enable react/no-unknown-property */}
 </Canvas>
+
+### External link
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="External link">
+    {() => (
```
robbieatacato, line 135: This is not necessary anymore: {() => {

robbieatacato (approved):

components/Button/src/index.scss
```diff
@@ -4,7 +4,7 @@
   border: 0;
   border-radius: var(--denhaag-button-border-radius);
   color: var(--denhaag-button-primary-action-color);
-  cursor: pointer;
+  cursor: default;
```
bddjong, line 7: Do we have a source or explanation for why this is better from accessibility pov? Testing this in storybook I feel like it doesn't convey properly that it's a clickable element.

bddjong (changes requested):

components/Button/src/index.scss
paulAcato: I'll check this with Ferdinand, not sure why 'default' is better than 'pointer' in any case for a button.

paulAcato: Asked the following to Fernand:

> Most of the buttons are simply links (<a href="[url]" class="denhaag-button"> ), should these also be given the default cursor? Or does this only apply to the situation <button class="denhaag-button"> ? What do you recommend in this?

And now we wait for a reply ;-) ( GDH-392 )

paulAcato: Hi @bddjong , i've received response of Fernand. 

When it's a link with the button class it must show the pointer, otherwise it must show the default cursor.

components/Button/src/stories/bem.stories.mdx
```diff
@@ -42,11 +40,9 @@ import readme from "../../README.md";
 <Canvas>
   {/* eslint-disable react/no-unknown-property */}
   <Story name="Secondary action">
-    {() => (
-      <div>
-        <button class={"denhaag-button denhaag-button--secondary-action"}>Button</button>
-      </div>
-    )}
+    <div>
+      <button class={"denhaag-button denhaag-button--secondary-action"}>Button</button>
```
robbieatacato, line 44: idem

components/Button/src/stories/bem.stories.mdx
```diff
@@ -56,26 +52,24 @@ import readme from "../../README.md";
...
-              <path d="M11.707 18.707a1 1 0 01-1.414 0l-6-6a1 1 0 010-1.414l6-6a1 1 0 111.414 1.414L7.414 11H19a1 1 0 110 2H7.414l4.293 4.293a1 1 0 010 1.414z"></path>
-            </svg>
-          </span>
-          Button
-        </button>
-      </div>
-    )}
+    <div>
+      <button class={"denhaag-button denhaag-button--secondary-action denhaag-button--start-icon"}>
+        <span class={"denhaag-button__icon"}>
```
robbieatacato, line 57: without the {}

components/Button/src/stories/bem.stories.mdx
```diff
@@ -116,13 +106,26 @@ import readme from "../../README.md";
...
   <Story name="Disabled">
-    {() => (
-      <div>
-        <button disabled class={"denhaag-button"}>
-          Button
-        </button>
-      </div>
-    )}
+    <div>
+      <button disabled class={"denhaag-button"}>
```
robbieatacato, line 110: without the {}

components/Button/src/stories/bem.stories.mdx
```diff
@@ -85,28 +79,24 @@ import readme from "../../README.md";
...
-              aria-hidden="true"
-            >
-              <path d="M12.293 5.293a1 1 0 011.414 0l6 6a1 1 0 010 1.414l-6 6a1 1 0 01-1.414-1.414L16.586 13H5a1 1 0 110-2h11.586l-4.293-4.293a1 1 0 010-1.414z"></path>
-            </svg>
-          </span>
-        </button>
-      </div>
-    )}
+    <div>
+      <button class={"denhaag-button denhaag-button--large denhaag-button--secondary-action denhaag-button--end-icon"}>
```
robbieatacato, line 83: without the {}

robbieatacato (changes requested):

PR #916, opened 2022-04-25
Refactor/pds component

bddjong:
**Closing issues**

PR #939, opened 2022-05-02
feat: more design tokens for Process Steps component

Robbert:
To make Process Steps work for the design of Gemeente Utrecht we needed some tweaks to the design tokens. This PR is not fully done yet, but considering the remarks in PR #916 about the necessity of further refactorings, I though I better share my work in progress early on.

### DIFF
--- a/components/ProcessSteps/src/Step.scss
+++ b/components/ProcessSteps/src/Step.scss
@@ -9,7 +9,7 @@
 }
 
 .denhaag-process-steps__step-line,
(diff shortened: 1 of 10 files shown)


Robbert:
Update: ready for review. The component for Den Haag still looks the same as before the refactoring and introduction of new tokens, but the Gemeente Utrecht version looks much different. One of the new options is to configure the size of the _sub step marker_.

Den Haag:

Screenshot 2022-05-02 at 14 39 40

Utrecht:

Screenshot 2022-05-02 at 14 40 58

PR #954, opened 2022-05-11
refactor(accordion): refactor for screen readers accessibility

robbieatacato:
Solve: https://acato-nl.atlassian.net/browse/GDH-445

Purpose:
- make accordion screen reader accessible

closes #954


paulAcato (approved):

bddjong:
@robbieatacato maybe we can improve this even further by adding the `aria-controls` attribute? 

See this [WAI-ARIA example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/accordion/accordion.html) (source code further down the page)

PR #958, opened 2022-05-12
feat: add posttypelabel react component

bddjong:
Adds a react export of the PostTypeLabel component


components/PosttypeLabel/src/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { HTMLAttributes } from 'react';
+import clsx from 'clsx';
+
+export const PostTypeLabel = (props: HTMLAttributes<HTMLParagraphElement>) => {
+  return (
+    <p {...props} className={clsx('denhaag-posttypelabel')}>
```
mjcarsjens, line 6: Might want to add the props.className here as well

PR #970, opened 2022-05-23
build: provide CommonJS exports in addition to ES Modules

### LINKED ISSUE #942: [Bug] Cannot use import statement outside a module
Via [Chantalle den Hertog @ GitLab](https://gitlab.com/chantalle_denhertog) working on [this issue](https://gitlab.com/commonground/huishoudboekje/app-new/-/issues/921):

---

# Enable testing a React application with NL Design System Den Haag components

### User story
As a developer I want to be able to use react-testing-library and jest to test my application that uses NL Design System Den Haag components.

### Additional context
At the moment I am using NLDS Den Haag components in my application. Once I try to run some test I get an error with all the components I am importing from the NLDS Den Haag. I've attached a screenshot of the errors that occur.

When I use components only from the NL Design System Utrecht I have no issues and the tests run fine. Therefor I think there seems to be an issue with the Den Haag components.

afbeelding

Yolijn:
As discussed in #942 and tested with PR https://github.com/nl-design-system/denhaag/pull/918 this PR results in both CommonJS exports and ESJ modules.

I've tested it with our example-with-next form and after these fixes I don't have the "Cannot use import statement outside a module" error anymore while including the FormProgress

**Closing issues**

closes #942


Robbert (approved):
looks good to me!

PR #972, opened 2022-05-24
Feature/gdh 408 edit links

paulAcato:
Applied HTML of denhaag-link to denhaag-list component. Therefor we could remove duplicate .denhaag-link styles.

Also added some stories for external links.

Refactor of link-styles, which were colliding (also due to duplicate application with @extend).


paulAcato:
Closed, had a call to check logic with Design. Will open new PR ;-)

PR #973, opened 2022-05-24
Feature/gdh 408 edit links

paulAcato:
- Edited stories, set new icon sizes with 24x24 pixels;
- Set styles with display flex and gap instead position absolute;
- Remove duplication due to .extend denhaag-link;


paulAcato:
@bddjong , tests fail due to screenshot comparison. I've changed the icon-size and output with design. There's a visual-change. How do we let it pass this check?

bddjong (approved):
Approved and updated the tests. Only note that the icon is 1px smaller than the figma design on both width and height. Once the CI finishes it can be merged to main.

PR #979, opened 2022-06-01
Feature/floating contact

paulAcato:
The Floating contact template for the HTML and CSS for the WordPress block.

Sadly for the LiveCom part isn't working jet. Therefor we're taking the risk that this must be completely refactored. Yesterday we contacted LiveCom again and kinda received the answer: "You have to implement it different" without mentioning how or providing any documents..


robbieatacato (commented):
The readme is empty.

components/FloatingContact/src/FloatingContact.js
```diff
@@ -0,0 +1,43 @@
...
+    this.links = [...this.element.getElementsByClassName(`${id}__item-link`)];
+
+    this.toggle.onclick = () => this.toggleEvents();
+  }
+
+  /**
+   * Toggle event. Fires onclick of the label.
+   */
+  toggleEvents() {
+    const open = this.toggle.getAttribute('aria-expanded') === 'false';
```
robbieatacato, line 38: I think you meant: `this.toggle.getAttribute('aria-expanded') === 'true';`

components/FloatingContact/src/index.tsx
```diff
@@ -0,0 +1 @@
+import './index.scss';
```
robbieatacato, line 1: This import is already in the bem file.
The import can be removed.

robbieatacato (changes requested):

components/FloatingContact/src/FloatingContact.js
paulAcato: No, i've checked the states again. toggleEvents is fired onclick of the button. We want the current-state before updating it.

Will rename it to `state`

robbieatacato (approved):

PR #989, opened 2022-06-14
style(link-group): fix arrow allignment

robbieatacato:
Ticket: https://acato-nl.atlassian.net/browse/GDH-610

Solving:

- vertical alignment of arrows
- horizontal space between arrow and text
- added html class to text span (linke the link component)

closes

### DIFF
--- a/proprietary/Components/src/denhaag/link-group.tokens.json
+++ b/proprietary/Components/src/denhaag/link-group.tokens.json
@@ -34,10 +34,11 @@
         "padding-block-end": { "value": "0" }
       },
       "denhaag-link-icon": {
-        "font-size": { "value": "0.75rem" }
+        "font-size": { "value": "0.75rem" },
+        "align-self": { "value": "start" },
+        "flex": { "value": "1 0 auto" }
       },
       "denhaag-link-icon-denhaag-icon": {
-        "align-self": { "value": "start" },
         "padding-block-start": { "value": "{denhaag.space.block.3xs}" }
       }
     }
--- a/components/LinkGroup/src/index.scss
+++ b/components/LinkGroup/src/index.scss
@@ -50,10 +50,11 @@
 
 .denhaag-link-group .denhaag-link__icon {
   font-size: var(--denhaag-link-group-denhaag-link-icon-font-size);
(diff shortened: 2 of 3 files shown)


proprietary/Components/src/denhaag/link-group.tokens.json
```diff
@@ -34,10 +34,11 @@
         "padding-block-end": { "value": "0" }
       },
       "denhaag-link-icon": {
-        "font-size": { "value": "0.75rem" }
+        "font-size": { "value": "0.75rem" },
+        "align-self": { "value": "start" },
+        "flex": { "value": "1 0 auto" }
```
bddjong, line 39: Do we really need these two values to be customisable through design tokens?

bddjong (changes requested):

proprietary/Components/src/denhaag/link-group.tokens.json
paulAcato: @robbieatacato , thought we discussed this would be applied on the denhaag-link instead of the link-group since all other elements require the denhaag-link component. 

@bddjong , the properties robbie added fixes the wierd icon alignment, where gap / icon width is reduced.

paulAcato (changes requested):
zie comment.

proprietary/Components/src/denhaag/link-group.tokens.json
bddjong: Can these properties not just be added to the css? Bit strange to require a layout fix to be added to the theme definition.

paulAcato: For the "flex: 1 0 auto;" I agree with you.

The alignment of the icons is in a lot of cases 'flex-start', but in several cases it must be 'center'. Therefor a CSS variable is in my opinion a good practice.

robbieatacato: I restored the tokens + scss.
Only left changes: the bem file with added `class="denhaag-link__label"` tot the span holding the text as the link component.

I wil move the styling changes to the link component.

PR #990, opened 2022-06-15
fix(floating contact): added keyboard events

robbieatacato:
Ticket: https://acato-nl.atlassian.net/browse/GDH-58

Solve

When the foldout is open, close it: 
- If you tab out of the fold-out (so if you tap away from 'info@denhaag.nl'); 
- If you shift-tab from the 'Contact' button; 
- With ESC.

https://user-images.githubusercontent.com/95216123/173815058-49af1ad1-100c-4a84-9825-d69237e36128.mov

closes #990

### DIFF
--- a/components/FloatingContact/src/stories/bem.stories.mdx
+++ b/components/FloatingContact/src/stories/bem.stories.mdx
@@ -54,7 +54,7 @@ import { useEffect, useMemo } from "@storybook/client-api";
           class="denhaag-floating-contact-switch"
           aria-expanded="false"
           data-label-open="Contact"
-          data-label-close="Sluiten"
+          data-label-close="Contact"
         >
           <div id="denhaag-floating-contact__label" class="denhaag-floating-contact-switch__label">
             Contact
@@ -210,7 +210,7 @@ import { useEffect, useMemo } from "@storybook/client-api";
           class="denhaag-floating-contact-switch denhaag-floating-contact-switch--hover"
           aria-expanded="false"
           data-label-open="Contact"
-          data-label-close="Sluiten"
+          data-label-close="Contact"
         >
           <div id="denhaag-floating-contact__label" class="denhaag-floating-contact-switch__label">
             Contact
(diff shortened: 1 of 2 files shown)


components/FloatingContact/src/stories/bem.stories.mdx
```diff
@@ -54,7 +54,7 @@ import { useEffect, useMemo } from "@storybook/client-api";
           class="denhaag-floating-contact-switch"
           aria-expanded="false"
           data-label-open="Contact"
-          data-label-close="Sluiten"
+          data-label-close="Contact"
```
paulAcato, line 57: I added the data-attributes `data-label-open` and `data-label-close` to set the readable label.
If this shouldn't  be used, please set the label inline and remove the JS that sets the label as innerText/innerHTML.

Also I'm wondering, the label for screenreaders must stay `contact`, but what about the toggle label? Shouldn't that still be "Close/Sluiten" for deskop users (without screenreaders). That said, an aria-label is probably better to set? Not sure, please check with Accessibility Expert and Design.

paulAcato (changes requested):
Check with Accessibility expert and design. Now you're replacing with JS "Contact" with "Contact"

components/FloatingContact/src/stories/bem.stories.mdx
robbieatacato: @paulAcato Fernands request: https://acato-nl.atlassian.net/browse/GDH-58?focusedCommentId=89613

paulAcato: I know that, but that isn't what I asked.

Now you're appyling by JS a label thats already there.

What I'm asking is if it's possible with an aria-label or other attribute so that desktop users **without** screenreaders see a normal text like: "Close".

Otherwise the JS toggle can be removed with those data-attributes and set the label hardcoded.

paulAcato (approved):

PR #991, opened 2022-06-15
style(link-with-icon): add arrow alignment link icon in list item

robbieatacato:
Ticket: https://acato-nl.atlassian.net/browse/GDH-610

Solving:

- vertical alignment of arrows in a list item
- horizontal space between arrow and text in a list item

Explanation:
I put it in a seperate list selector, because putting the css properties directly into the `.denhaag-link__icon` gave the vertical alignment of the link in the paragraph a different vertical alignment.

Screenshot 2022-06-15 at 11 01 06
Screenshot 2022-06-15 at 11 01 36

closes #991

### DIFF
--- a/components/Link/src/index.scss
+++ b/components/Link/src/index.scss
@@ -71,7 +71,7 @@
   width: var(--denhaag-link-icon-size);
 }
 
-li .denhaag-link--with-icon .denhaag-link__icon {
+li .denhaag-link__icon {
   align-self: start;
   flex: 1 0 auto;
 }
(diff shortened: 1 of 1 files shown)


components/Link/src/index.scss
```diff
@@ -71,6 +71,11 @@
   width: var(--denhaag-link-icon-size);
 }
 
+li .denhaag-link--with-icon-start .denhaag-link__icon {
+  align-self: start;
```
paulAcato, line 75: Uhh, this is to specific, since we also can align icons on the end of the line.

Therefor i would suggest again to replace the `align-items:` with `align-self` on line 64 of this file. The flex: 1 0 auto may also be appended to the `.denhaag-link__icon` class.

paulAcato (changes requested):

paulAcato (approved):

PR #1027, opened 2022-07-19
refactor(anchor-navigation): change aria labels, remove ids ol for ac…

robbieatacato:
Solve: accessibility (https://acato-nl.atlassian.net/browse/GDH-448)

Purpose:

- accessibility screen readers (adviced by Fernand van Olphen)

closes #1027


Robbert (approved):

Yolijn (approved):
Looks good, makes more sense than the `aria-labelledby` which pointed to a list :) 
As this is only a change in the Storybook, is this the template you'll also be using in the Gutenberg blocks for this component?

PR #1031, opened 2022-07-21
refactor(cta-image-content): change figure + figcaption

robbieatacato:
### Solve: 
--Quote Fernand van Olphen (accessibility The Hague):
Figure and figcaption are elements that are not necessary for accessibility.
We have decided not to give the image in this component alternative text.
A figcaption is an element that is used to complement the figure element.
It says something extra about what is in the figure element.
But since the image has no alternative text, the figcaption complements... nothing.

### Purpose:

- accessibility

closes #1031


paulAcato (approved):
Looking good to me

PR #1041, opened 2022-08-01
style(accordion): styling corrections panel titles

robbieatacato:
### Solve: 
wrong/not up-to-date styling https://acato-nl.atlassian.net/browse/GDH-665

### Purpose: 
correct font-size, font-weight, font-family and color for accorion panel titles

- closed and opened: font-size 18px, font-family TheSans
- when closed: default (font-weight 400), hover (font-weight 400, color green 3) and focus (font-weight 400, color green 3)
- when opened: default (font-weight 700 ,color green 3), hover (font-weight 700, color green 4) and focus (font-weight 700, color green 3)

### EXTRA: 

- overhaul/refactor of bem, css and tokens.
- added 'open', 'open - hover' and 'open - focus' bem for more clearity in styling

### Figma: 
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1585%3A0

closes #1041

### DIFF
--- a/components/Accordion/src/index.scss
+++ b/components/Accordion/src/index.scss
@@ -25,6 +25,10 @@
 }
 
 .denhaag-accordion__panel {
+  position: inherit;
+  width: inherit;
+  font-family: inherit;
+  font-weight: inherit;
   outline: var(--denhaag-accordion-panel-outline);
   align-items: var(--denhaag-accordion-panel-align-items);
   border-radius: var(--denhaag-accordion-panel-border-radius);
@@ -34,15 +38,14 @@
   padding-block-start: var(--denhaag-accordion-panel-padding-block);
   padding-block-end: var(--denhaag-accordion-panel-padding-block);
   padding-inline-end: var(--denhaag-accordion-panel-padding-inline);
-  position: relative;
   min-height: var(--denhaag-accordion-panel-min-height);
   margin-block-start: var(--denhaag-accordion-panel-margin-block);
   margin-block-end: var(--denhaag-accordion-panel-margin-block);
-  width: var(--denhaag-accordion-panel-width);
+  font-size: var(--denhaag-accordion-panel-font-size);
 }
 
-.denhaag-accordion__panel > .denhaag-icon {
-  color: var(--denhaag-accordion-icon-color);
+.denhaag-accordion__panel .denhaag-icon {
+  color: inherit;
   position: var(--denhaag-accordion-icon-position);
   top: var(--denhaag-accordion-icon-top);
   right: var(--denhaag-accordion-icon-right);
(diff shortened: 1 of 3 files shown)


components/Accordion/src/index.scss
```diff
@@ -52,6 +52,8 @@
 .denhaag-accordion__title {
   --utrecht-heading-5-margin-block-start: 0;
   --utrecht-heading-5-margin-block-end: 0;
+  --utrecht-heading-5-font-family: var(--denhaag-accordion-font-family);
+  --utrecht-heading-5-font-weight: 400;
```
paulAcato, line 56: Apply var(--denhaag-typography-weight-regular);  (not sure if this is the 400 value)

components/Accordion/src/index.scss
```diff
@@ -94,6 +96,7 @@
 
 .denhaag-accordion__title:focus-visible,
 .denhaag-accordion__title--focus {
+  color: var(--denhaag-accordion-panel-color);
```
paulAcato, line 99: Split to: --denhaag-accordion-panel-*focus-*color

components/Accordion/src/index.scss
```diff
@@ -117,18 +120,30 @@
...
+.denhaag-accordion__container--open .denhaag-accordion__title:not([disabled]) {
+  --utrecht-heading-5-color: var(--denhaag-accordion-panel-color);
+  --utrecht-heading-5-font-weight: 700;
+}
+
 .denhaag-accordion__panel:focus .denhaag-accordion__title:not([disabled]),
 .denhaag-accordion__panel:focus .denhaag-accordion__title:not([disabled]) + .denhaag-icon,
 .denhaag-accordion__panel:hover .denhaag-accordion__title:not([disabled]),
 .denhaag-accordion__panel:hover .denhaag-accordion__title:not([disabled]) + .denhaag-icon {
   color: var(--denhaag-accordion-panel-color);
```
paulAcato, line 136: Split to open color

paulAcato (changes requested):

components/Accordion/src/index.scss
Robbert, line 56: This styling of `heading-5` tokens seems like a very roundabout way to style the h3 of the accordion differently.

Seems to me there a much simpler and more flexible approach is possible: style the `<button>` to not have its own style, but to inherit font-family, color, etc. For example:

```css
.denhaag-accordion__title-button {
  color: inherit;
  font-family: inherit;
  font-size: inherit;
  line-height: inherit;
  width: 100%;
  /* etc */
}
```

components/Accordion/src/index.scss
```diff
@@ -117,18 +120,30 @@
...
 
-.denhaag-accordion__container--open .denhaag-accordion__panel:hover,
 .denhaag-accordion__container--open .denhaag-accordion__panel > .denhaag-icon:hover {
   --denhaag-accordion-title-color: var(--denhaag-accordion-container-open-color);
 }
 
+.denhaag-accordion__container--open .denhaag-accordion__title:not([disabled]) {
+  --utrecht-heading-5-color: var(--denhaag-accordion-panel-color);
+  --utrecht-heading-5-font-weight: 700;
+}
```
Robbert, line 130: If you want to force the heading of the accordion to have a certain style, and override the heading default styling, I suggest two options:

- do not use `utrecht-heading-*` at all, just use `<h2 class="denhaag-accordion_title">`-`<h6>`
- do not only style `--utrecht-heading-5`, but also all the tokens for `utrecht-heading-2`/`3`/`4`/`6`. Accordions can occur at any level in the page and should not be hardcoded to `h3` or `h5`. That would be a WCAG violation.

paulAcato: @Robbert , What about the element can change from heading-size but the CSS selector stays the same on utrecht-heading-5?

Robbert: That is fine too from an accessibility standpoint, but what would be the advantage? Still seems like an odd choice to use the `heading-5` element to style the accordion heading, instead of the other way around, if you want to apply some of the design choices from heading 5.

You could do:

```json
{
  "denhaag": {
    "accordion": {
      "title": { "font-size": { "value": "{utrecht.heading-5.font-size}" } }
    }
  }
}
```

paulAcato: The only advantage would be less bloating in duplicate styles

robbieatacato:
@Yolijn @Robbert @paulAcato Ik heb de gehele boel een overhaul gegeven:

- BEM refactor
- css refactor
- tokens refactor

components/Accordion/src/index.scss
```diff
@@ -34,43 +38,51 @@
...
   padding-inline-end: var(--denhaag-accordion-panel-padding-inline);
-  position: relative;
   min-height: var(--denhaag-accordion-panel-min-height);
   margin-block-start: var(--denhaag-accordion-panel-margin-block);
   margin-block-end: var(--denhaag-accordion-panel-margin-block);
-  width: var(--denhaag-accordion-panel-width);
+  font-size: var(--denhaag-accordion-panel-font-size);
 }
 
-.denhaag-accordion__panel > .denhaag-icon {
```
paulAcato, line 44: Leave this be. I've added this because this was causing an issue with other icons within the panels

paulAcato (changes requested):

components/Accordion/src/index.scss
robbieatacato: @paulAcato DONE

paulAcato (approved):

PR #1055, opened 2022-08-16
fix: remove flexbox logic from .denhaag-cta-link__excerpt

Yolijn:
to align it with the [Figma designs](https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=8105%3A21885)
and to prevent weird whitespace and wrap issues

### DIFF
--- a/components/CtaLink/src/index.scss
+++ b/components/CtaLink/src/index.scss
@@ -16,7 +16,6 @@
   padding-inline-start: var(--denhaag-cta-link-padding-inline, calc(var(--denhaag-cta-link-gap) / 1.5));
   padding-inline-end: var(--denhaag-cta-link-padding-inline, calc(var(--denhaag-cta-link-gap) / 1.5));
   text-decoration: none;
-  word-break: break-word;
 }
 
 .denhaag-cta-link:hover,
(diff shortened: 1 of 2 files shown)


components/CtaLink/src/index.scss
```diff
@@ -16,7 +16,6 @@
   padding-inline-start: var(--denhaag-cta-link-padding-inline, calc(var(--denhaag-cta-link-gap) / 1.5));
   padding-inline-end: var(--denhaag-cta-link-padding-inline, calc(var(--denhaag-cta-link-gap) / 1.5));
   text-decoration: none;
-  word-break: break-word;
```
Yolijn, line 19: Screenshot 2022-08-16 at 15 03 51

The `break-word` option was causing the words to be unreadable when the screen was small enough to actually break them.

@paulAcato I believe you've added this before to solve a problem, do you remember what it solved so we can ensure this will not reintroduce it?

components/CtaLink/src/stories/bem.stories.mdx
```diff
@@ -46,9 +46,9 @@ import readme from "../../README.md";
           />
         </svg>
       </div>
-      <div class="denhaag-cta-link__excerpt">
+      <p class="denhaag-cta-link__excerpt">
```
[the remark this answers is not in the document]
paulAcato: In HTML 4 or HTMLX this isn't allowed. HTML5 allows this, but sometimes it's still flaky (issue with inline and block elements). Therefor are we taking the risk?

components/CtaLink/src/index.scss
paulAcato: This can be removed. Can't find anymore why this was added.

paulAcato (approved):

PR #1067, opened 2022-08-24
style(button): additional bem

robbieatacato:
### Solve:

Ticket: https://acato-nl.atlassian.net/browse/GDH-470

Figma: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=1%3A656

- to view bem variations for default and large buttons: icon left, icon right
- to view bem states for default and large buttons: hover, disabled, focus

### Purpose:

- add additional bem variations for default and large buttons: icon left, icon right
- add additional bem states for default and large button: hover, disabled, focus

closes

### DIFF
--- a/components/Button/src/stories/bem.stories.mdx
+++ b/components/Button/src/stories/bem.stories.mdx
@@ -29,25 +29,36 @@ import readme from "../../README.md";
   {/* eslint-disable react/no-unknown-property */}
   <Story name="Default">
     <button class="denhaag-button">Button</button>
-  </Story>
-  {/* eslint-enable react/no-unknown-property */}
-</Canvas>
-
-### Secondary action
-
-<Canvas>
-  {/* eslint-disable react/no-unknown-property */}
-  <Story name="Secondary action">
+    <br />
+    <br />
     <button class="denhaag-button denhaag-button--secondary-action">Button</button>
   </Story>
   {/* eslint-enable react/no-unknown-property */}
 </Canvas>
 
-### Icon left
+### Default - icon left
 
 <Canvas>
   {/* eslint-disable react/no-unknown-property */}
-  <Story name="Icon left">
+  <Story name="Default - icon left">
+    <button class="denhaag-button denhaag-button--start-icon">
+      <span class="denhaag-button__icon">
+        <svg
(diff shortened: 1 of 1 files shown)


components/Button/src/stories/bem.stories.mdx
```diff
@@ -29,26 +29,319 @@ import readme from "../../README.md";
   {/* eslint-disable react/no-unknown-property */}
   <Story name="Default">
     <button class="denhaag-button">Button</button>
+    <br />
```
paulAcato, line 32: Why the `<br />` ? This is shown to the one who tries to read the documentation. This must be as clean as possible.

I would rather suggest to add a storybook.css file you import which includes something like `.denhaag-button + .denhaag-button { margin-block-start: 1rem; }`

components/Button/src/stories/bem.stories.mdx
```diff
@@ -29,26 +29,319 @@ import readme from "../../README.md";
...
+
+<Canvas>
+  {/* eslint-disable react/no-unknown-property */}
+  <Story name="Default - hover">
+    <button class="denhaag-button denhaag-button--hover">Button</button>
+    <br />
+    <br />
+    <button class="denhaag-button denhaag-button--secondary-action denhaag-button--hover">Button</button>
+    <br />
+    <br />
```
paulAcato, line 135: idem

paulAcato (changes requested):
I'm missing the style changes?

components/Button/src/stories/bem.stories.mdx
robbieatacato: @paulAcato the <br /> are purely for storybook, not for styling

paulAcato: Yea i know that. But the Storybook canvas must only show what it needs to show. 

There's also a button-group that adds spacing between buttons. This solution is quick and ugly. If a new developer (or someone without much HTML knowledge)  opens it and sees 3 to 10 `<br/>` will get quite confused.

paulAcato (approved):

PR #1068, opened 2022-08-25
feat: added login page template + all needed CSS/HTML components

YourivHDenHaag:
Work in progress: adding CTA Card (for login pages)


components/Header/src/index.scss
```diff
@@ -0,0 +1,10 @@
+.denhaag-header {
+  border-bottom: var(--denhaag-header-border-bottom);
```
Yolijn, line 2: If we use `--denhaag-header-border-bottom-color`, `--denhaag-header-border-bottom-width` etc. and a hardcoded `border-bottom-style: solid;` we will be able to use the same key/value pairs in Figma as this doesn't support the same composite tokens. The same goes for a hypothetical other output such as for iOS or Android. :)

components/Header/src/index.scss
```diff
@@ -0,0 +1,10 @@
...
+
+  &__container {
```
Yolijn, line 4: Seems odd to call the content of a container (the header in this case) container. Could we maybe call it `__content`?

components/Page/src/index.scss
```diff
@@ -0,0 +1,70 @@
+.denhaag-page {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr 1fr;
```
Yolijn, line 3: Can we use `grid-template-columns: repeat(12, 1fr);` instead?

components/Page/src/index.scss
```diff
@@ -0,0 +1,70 @@
...
+  grid-auto-flow: column;
+
+  &-header {
+    grid-column: 1 / 13;
+  }
+
+  /* Den Haag XXS */
+  @media (max-width: 359px) {
```
Yolijn, line 11: Lets use CSS variables for all of it, based on the Design Tokens in Proprietary

```css
/* From the proprietary/components/denhaag/page.tokens.json file */
.denhaag-theme {
    --denhaag-page-grid-column-gap: 8px;
    --denhaag-page-grid-margin-inline-start: 24px;
    --denhaag-page-grid-margin-inline-end: 24px;
    --denhaag-page-xs-grid-column-gap: 8px;
    --denhaag-page-xs-grid-margin-inline-start: 24px;
    --denhaag-page-xs-grid-margin-inline-end: 24px;
}

.denhaag-page {
     grid-column-gap: var(--denhaag-page-grid-column-gap);
     margin-inline-start: var(--denhaag-page-margin-inline-start);
     margin-inline-end: var(--denhaag-page-margin-inline-end);
    
    @media (min-width: 360px) {
        --denhaag-page-grid-column-gap: var(--denhaag-page-xs-grid-column-gap, var(--denhaag-page-grid-column-gap));
        --denhaag-page-grid-margin-inline-start: var(--denhaag-page-grid-xs-margin-inline-start, var(--denhaag-page-grid-margin-inline-start));
        --denhaag-page-grid-margin-inline-end: var(--denhaag-page-grid-margin-inline-end, var(--denhaag-page-grid-margin-inline-end));
    }

    @media (min-width: 768px) {
        ...
    }
}
```

.storybook/stories/templates/Login-page.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+import Breadcrumb from "/components/Breadcrumb/src";
+import { AuthenticationCardGroup } from "/components/CardGroup/src/story-template";
+
+import clsx from "clsx";
+
+import "./story-login.css";
+
+<Meta title="Templates/Login page" />
+
+# Summary page
```
Yolijn, line 19: Hmmm is het eigenlijk de Summary page of the Login page?

proprietary/Components/src/denhaag/responsive-content.tokens.json
```diff
@@ -0,0 +1,19 @@
+{
+  "denhaag": {
+    "responsive-content": {
+      "max-width": { "value": "calc(100vw - (var(--denhaag-space-inline-xl)))" },
```
Yolijn, line 4: Unfortunately this `calc` is not something that is supported outside of `css` and therefor not a great thing to use in the Style Dictionary design tokens. So we should use the actual value instead :(

Yolijn:
@YourivHDenHaag I think we're almost there. I left two comments and I think mostly the `calc` in style-dictionary should be resolved

.storybook/stories/templates/Login-page.stories.mdx
YourivHDenHaag: Login page, i've changed it.

proprietary/Components/src/denhaag/responsive-content.tokens.json
YourivHDenHaag: I've removed the calc function and added a border-box alternative with a padding design token.

Yolijn (approved):

PR #1071, opened 2022-08-26
style(ol and ul): corrections + additional bem

robbieatacato:
### Solve: https://acato-nl.atlassian.net/browse/GDH-810

Figma: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=8105%3A22880

- wrong ordered -and unorderded list stylings
- styling issues when there is no class in a sub  `<ul>`
- styling issues when there is an paragraph inside a list item

### Purpose:

- style corrections
- additional style fallbacks
- additional bem for situations where there are paragraphs  inside the list items (for any reasons)
- remove alphabetical list items inside an unordered list (this was not requested by UX and not applied in figma)

closes #1071

### DIFF
--- a/components/OrderedList/src/index.scss
+++ b/components/OrderedList/src/index.scss
@@ -4,6 +4,7 @@
   @extend .denhaag-unordered-list;
 }
 
-.denhaag-ordered-list .denhaag-unordered-list {
+.denhaag-ordered-list .denhaag-unordered-list,
+.denhaag-ordered-list ul {
   @extend .denhaag-unordered-list--lower-alpha;
 }
--- a/components/OrderedList/src/index.scss
+++ b/components/OrderedList/src/index.scss
@@ -4,7 +4,6 @@
   @extend .denhaag-unordered-list;
 }
 
-.denhaag-ordered-list .denhaag-unordered-list,
-.denhaag-ordered-list ul {
+.denhaag-ordered-list .denhaag-unordered-list {
   @extend .denhaag-unordered-list--lower-alpha;
 }
--- a/components/UnorderedList/src/index.scss
+++ b/components/UnorderedList/src/index.scss
@@ -22,3 +22,28 @@
 .denhaag-unordered-list--lower-alpha {
   list-style-type: var(--denhaag-unordered-list-lower-alpha-list-style-type);
 }
+
+.denhaag-unordered-list ul {
+  @extend .denhaag-unordered-list;
+
+  margin-inline-start: initial;
+  margin-inline-end: initial;
+  margin-block-start: initial;
+  margin-block-end: initial;
+}
+
+.denhaag-unordered-list li {
+  line-height: var(--denhaag-unordered-list-line-height, 1.5);
+}
+
+.denhaag-unordered-list li > p {
+  line-height: inherit;
+  margin-inline-start: inherit;
+  margin-inline-end: inherit;
+  margin-block-start: inherit;
+  margin-block-end: inherit;
+  padding-inline-start: inherit;
+  padding-inline-end: inherit;
+  padding-block-start: inherit;
+  padding-block-end: inherit;
+}
--- a/components/UnorderedList/src/index.scss
+++ b/components/UnorderedList/src/index.scss
@@ -22,28 +22,3 @@
 .denhaag-unordered-list--lower-alpha {
   list-style-type: var(--denhaag-unordered-list-lower-alpha-list-style-type);
 }
-
-.denhaag-unordered-list ul {
-  @extend .denhaag-unordered-list;
-
-  margin-inline-start: initial;
-  margin-inline-end: initial;
-  margin-block-start: initial;
-  margin-block-end: initial;
-}
-
-.denhaag-unordered-list li {
-  line-height: var(--denhaag-unordered-list-line-height, 1.5);
-}
-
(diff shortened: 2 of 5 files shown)


paulAcato (approved):
Code approved.

For those who still thought a Paragraphs aren't allowed in list-items:

`For semantic purposes, content within the li element should be wrapped in a p tag. If there is more than one paragraph or section in a li, do not close the li until after the last element. This differs from the current standard, but is cleaner for accessible content.`

components/UnorderedList/src/index.scss
```diff
@@ -22,3 +22,28 @@
 .denhaag-unordered-list--lower-alpha {
   list-style-type: var(--denhaag-unordered-list-lower-alpha-list-style-type);
 }
+
+.denhaag-unordered-list ul {
```
Yolijn, line 26: Why do we need `ul` in this selector? Isn't `denhaag-unordered-list` a class name on the `ul`?

Yolijn: > If for what reason, the nested `<ul>` doesn't have class="denhaag-unordered-list" this still will have the same css properties as .denhaag-unordered-list

Yeah thats exactly what I think you do not always want. I am wondering if we should have a wrapper component where such behaviour is added, maybe even for a lot more rich-text components without a class name. A bit like the [html-content component of Utrecht](https://github.com/nl-design-system/utrecht/blob/main/components/html-content/css/index.scss) 

Would also help quite a lot for unexpected behaviour of whitespace between components that is helpful and expected in the context of the (wordpress) website, but not necessarily within the context of other applications.

components/UnorderedList/src/index.scss
```diff
@@ -22,3 +22,28 @@
...
+  margin-inline-end: initial;
+  margin-block-start: initial;
+  margin-block-end: initial;
+}
+
+.denhaag-unordered-list li {
+  line-height: var(--denhaag-unordered-list-line-height, 1.5);
+}
+
+.denhaag-unordered-list li > p {
```
Yolijn, line 39: I would very much avoid directly selecting html elements instead of BEM class names. This behaviour might be consistently needed for the website but will cause unexpected behaviour for others. Please use `utrecht-unordered-list__item` or `denhaag-unordered-list__item` and `utrecht-paragraph` instead.

components/UnorderedList/src/index.scss
```diff
@@ -22,3 +22,28 @@
...
+  line-height: inherit;
+  margin-inline-start: inherit;
+  margin-inline-end: inherit;
+  margin-block-start: inherit;
+  margin-block-end: inherit;
+  padding-inline-start: inherit;
+  padding-inline-end: inherit;
+  padding-block-start: inherit;
+  padding-block-end: inherit;
```
Yolijn, line 48: I'm also wondering why it's needed to set all these properties to `inherit` what issue is it solving?

Yolijn (changes requested):

components/UnorderedList/src/index.scss
Robbert: > this is meant for a nested <ul> without the class="denhaag-unordered-list".
> For fallback purpose.

I don't really see who would benefit from this currently. Doesn't WordPress output class names for nested list items? The Gutenberg blocks appear to add those class names, from the looks of it.

components/UnorderedList/src/index.scss
```diff
@@ -22,3 +22,28 @@
...
+.denhaag-unordered-list li {
+  line-height: var(--denhaag-unordered-list-line-height, 1.5);
+}
+
+.denhaag-unordered-list li > p {
+  line-height: inherit;
+  margin-inline-start: inherit;
+  margin-inline-end: inherit;
+  margin-block-start: inherit;
+  margin-block-end: inherit;
```
Robbert, line 44: Isn't the intention to remove the margin only from the first and the last paragraph, but not the margins between paragraphs in list items?

components/UnorderedList/src/index.scss
robbieatacato: @Yolijn @Robbert I removed the fallback css now

components/UnorderedList/src/index.scss
robbieatacato: I removed the fallback css now

components/UnorderedList/src/index.scss
robbieatacato: I removed the fallback css now

components/UnorderedList/src/index.scss
robbieatacato: I removed the fallback css now

components/OrderedList/src/index.scss
```diff
@@ -4,6 +4,7 @@
   @extend .denhaag-unordered-list;
 }
 
-.denhaag-ordered-list .denhaag-unordered-list {
+.denhaag-ordered-list .denhaag-unordered-list,
+.denhaag-ordered-list ul {
```
paulAcato, line 8: If all fallback must be removed, this `ul` must be removed aswel

paulAcato (changes requested):

components/OrderedList/src/index.scss
robbieatacato: @paulAcato done, removed it

PR #1079, opened 2022-08-30
refactor: rename and simplify authenticationblock

Yolijn:
Rename AuthenticationBlock to CardAuthentication and move responsive logic to CardGroup component

- [x] Use design tokens
- [x] BEM all the things
- [x] Use mixins for the different styles and include them in both a modifier class and a responsive viewport
- [x] Use utility CSS custom property so a fallback value is always used and will stay available
- [x] Add stories for Cards in normal and portrait mode
- [x] Add stories for CardGroup with Authentication Card for all situations


.stylelintrc.json
```diff
@@ -8,7 +8,7 @@
     "scss/percent-placeholder-pattern": "^(denhaag|utrecht)-[a-z0-9-]+$",
     "scss/operator-no-newline-after": null,
     "scss/at-extend-no-missing-placeholder": null,
-    "custom-property-pattern": "^(denhaag|utrecht)-[a-z0-9-]+$",
+    "custom-property-pattern": "^(denhaag|_denhaag|utrecht)-[a-z0-9-]+$",
```
Robbert, line 11: How about "optional underscore" `_?`

 `"^_?(denhaag|utrecht)-[a-z0-9-]+$",`
 
 That would make it possible to extend `utrecht-button` and specify `--_utrecht-button-background-color`.

components/CardGroup/src/story-template.tsx
```diff
@@ -0,0 +1,136 @@
...
+    DigiD: '/digid.svg',
+    eHerkenning: '/eherkenning.png',
+    eIDAS: '/eidas.svg',
+  };
+
+  return (
+    <section className={clsx('denhaag-card-authentication', portrait && 'denhaag-card-authentication--portrait')}>
+      <header className="denhaag-card-authentication__header">
+        <hgroup className="denhaag-card-authentication__heading">
+          <h3>{title}</h3>
```
Robbert, line 38: Should this `h3` have `class="utrecht-heading-3"`?

.stylelintrc.json
Yolijn: Nice one 👍

PR #1088, opened 2022-09-09
Style/gdh 856 floating contact

robbieatacato:
### Solve:

- after some reasearch by The Hague UX team, they conlcuded the floating contact button did not stand out enough

### Purpose:

- change blue color scheme to a green one, so it is more standing out /visible
- added some extra spacing on the right and bottom of button for desktop

### Figma:
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=11566%3A65912

closes #1088

### DIFF
--- a/components/FloatingContact/src/index.scss
+++ b/components/FloatingContact/src/index.scss
@@ -40,7 +40,7 @@
 }
 
 @media (min-width: 1024px) {
-  .denhaag-floating-contact {
+  .denhaag-theme {
     --denhaag-floating-contact-bottom: var(--denhaag-space-block-2xl);
     --denhaag-floating-contact-right: var(--denhaag-space-inline-2xl);
   }
--- a/components/FloatingContact/src/index.scss
+++ b/components/FloatingContact/src/index.scss
@@ -39,6 +39,13 @@
   z-index: var(--denhaag-floating-contact-index);
 }
 
+@media (min-width: 1024px) {
+  .denhaag-floating-contact {
+    --denhaag-floating-contact-bottom: 2rem;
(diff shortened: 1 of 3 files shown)


components/FloatingContact/src/index.scss
```diff
@@ -39,6 +39,13 @@
   z-index: var(--denhaag-floating-contact-index);
 }
 
+@media (min-width: 1024px) {
+  .denhaag-floating-contact {
+    --denhaag-floating-contact-bottom: 2rem;
```
paulAcato, line 44: I would rather see that the denhaag spacing token is used.

example: `--denhaag-floating-contact-bottom: var(--denhaag-space-block-2xl);`  and `--denhaag-floating-contact-right: var(--denhaag-space-inline-2xl);`

paulAcato (changes requested):
Just a small change ;-)

paulAcato (approved):

components/FloatingContact/src/index.scss
robbieatacato: Aangepast

components/FloatingContact/src/index.scss
```diff
@@ -39,6 +39,13 @@
   z-index: var(--denhaag-floating-contact-index);
 }
 
+@media (min-width: 1024px) {
+  .denhaag-floating-contact {
```
Robbert, line 43: `.denhaag-floating-contact` ➝ `.denhaag-theme`

robbieatacato: @Robbert renamed it.

PR #1092, opened 2022-09-09
style(typography): change font-family, mobile font-sizes

robbieatacato:
### Solve:

- headings on mobile are to big
- font-family for headings is not the correct one

### Purpose:

- add mobile headings font-sizes
- change headings font-family to theMix

### Figma:

- https://www.figma.com/file/fCCe3DA2gCfsqYOm0vuPsi/Final-UI---Transitie-Online?node-id=7416%3A91424

closes #1092

### DIFF
--- a/components/Typography/src/heading.scss
+++ b/components/Typography/src/heading.scss
@@ -19,13 +19,7 @@
 .utrecht-heading-6 {
   /* TODO: Move to container element, make spacing optional */
   --utrecht-space-around: 1;
-}
 
-.utrecht-heading-1,
-.utrecht-heading-2,
-.utrecht-heading-3,
-.utrecht-heading-4,
-.utrecht-heading-5 {
   font-family: var(--denhaag-typography-sans-serif-alternate);
 }
 
--- a/components/Typography/src/heading.scss
+++ b/components/Typography/src/heading.scss
@@ -20,17 +20,13 @@
   /* TODO: Move to container element, make spacing optional */
   --utrecht-space-around: 1;
 
-  font-family: var(--denhaag-typography-sans-serif-alternate);
+  font-family: var(--utrecht-heading-font-family);
 }
 
 @media (max-width: 768px) {
-  .utrecht-heading-1 {
-    --utrecht-heading-1-font-size: 2.25rem;
-  }
-  .utrecht-heading-2 {
-    --utrecht-heading-2-font-size: 1.75rem;
-  }
-  .utrecht-heading-3 {
-    --utrecht-heading-3-font-size: 1.4375rem;
+  .denhaag-theme {
+    --utrecht-heading-1-font-size: var(--denhaag-typography-mobile-scale-3xl-font-size);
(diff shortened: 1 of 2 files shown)


components/Typography/src/heading.scss
```diff
@@ -19,4 +19,18 @@
 .utrecht-heading-6 {
   /* TODO: Move to container element, make spacing optional */
   --utrecht-space-around: 1;
+
+  font-family: var(--denhaag-typography-sans-serif-alternate);
+}
```
Robbert, line 24: Wouldn't it instead be possible to specify `utrecht.heading-6.font-family` in the design tokens JSON?

components/Typography/src/heading.scss
```diff
@@ -19,4 +19,18 @@
...
+@media (max-width: 768px) {
+  .utrecht-heading-1 {
+    --utrecht-heading-1-font-size: 2.25rem;
+  }
+  .utrecht-heading-2 {
+    --utrecht-heading-2-font-size: 1.75rem;
+  }
+  .utrecht-heading-3 {
+    --utrecht-heading-3-font-size: 1.4375rem;
+  }
```
Robbert, line 35: I don't think you need to specify the component, you could manage the responsive theme in a separate file:

```css
@media (max-width: 768px) {
   .denhaag-theme {
     --utrecht-heading-1-font-size: 2.25rem;
     --utrecht-heading-2-font-size: 1.75rem;
     --utrecht-heading-3-font-size: 1.4375rem;
   }
}
```

By separating the theme from the component, it remains usable as white-label component for other teams.

Taking it one step further, I think it would be good to specify the entire mobile scale from Figma in `typography.tokens.json` so you can use:

```css
@media (max-width: 768px) {
   .denhaag-theme {
     --utrecht-heading-1-font-size: var(--denhaag-typography-mobile-scale-3xl-font-size);
     --utrecht-heading-2-font-size: var(--denhaag-typography-mobile-scale-2xl-font-size);
     --utrecht-heading-3-font-size: var(--denhaag-typography-mobile-scale-xl-font-size);
   }
}
```

Robbert (changes requested):

components/Typography/src/heading.scss
robbieatacato: @Robbert which tokens file? The utrecht or THe Hague

components/Typography/src/heading.scss
robbieatacato: @Robbert I applied your feadback into the newest commit.

Robbert: Nice! Eventually (in another PR) we will need to find a way to combine all these responsive `denhaag-theme` customizations in one file, so they can be shipped along with the design tokens. @Yolijn any ideas?

Yolijn: 🤔 Good idea @Robbert I don't have a thought out idea at the moment, but let's brainstorm about it this week!

robbieatacato:
@Robbert I handled your feedback I think.

Robbert (approved):

PR #1111, opened 2022-09-26
style(heading): responsive margins

robbieatacato:
### Solve:

- margin above headings headings 1, 2, 3, 4, 5 when following other components

### Purpose:

- apply repsonsive margins above headings 1, 2, 3, 4, 5 when following other components

### Figma:

- https://www.figma.com/file/fCCe3DA2gCfsqYOm0vuPsi/Final-UI---Transitie-Online?node-id=7445%3A94085

closes  #1111

### DIFF
--- a/components/Typography/src/heading.scss
+++ b/components/Typography/src/heading.scss
@@ -27,19 +27,23 @@
   --utrecht-heading-1-margin-block-start: var(--denhaag-space-block-xs, 0.5rem) !important;
 }
 
-* + .utrecht-heading-2 {
+*:not(.denhaag-columns):not(.denhaag-button-group):not(.denhaag-featured-posts):not(.denhaag-page-buttons):not(.denhaag-image)
+  + .utrecht-heading-2 {
   --utrecht-heading-2-margin-block-start: var(--denhaag-space-block-4xl);
 }
 
-* + .utrecht-heading-3 {
+*:not(.denhaag-columns):not(.denhaag-button-group):not(.denhaag-featured-posts):not(.denhaag-page-buttons):not(.denhaag-image)
+  + .utrecht-heading-3 {
   --utrecht-heading-3-margin-block-start: var(--denhaag-space-block-3xl);
 }
 
-* + .utrecht-heading-4 {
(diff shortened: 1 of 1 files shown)


components/Typography/src/heading.scss
```diff
@@ -22,10 +22,54 @@
   font-family: var(--utrecht-heading-font-family);
 }
 
+.utrecht-heading-1:first-child {
+  /* overwriting root :root .utrecht-heading-1:first-child */
+  --utrecht-heading-1-margin-block-start: var(--denhaag-space-block-xs, 0.5rem) !important;
+}
+
+* + .utrecht-heading-2 {
```
paulAcato, line 30: `*` is every component including the columns which already have a spacing. I would suggest here a list of `component:not(.component) +.utrecht-heading-X`.

Components like a button-group,  accordion-group, cta-link-group already have spacings around them, i think this will cause an issue.

Did you check how this would look in the WP environment on a page with all possible combinations?

paulAcato (changes requested):

components/Typography/src/heading.scss
robbieatacato: @paulAcato isn't that a wordpress theme specific exception?
I can put that in the theme itself like this: `*:not(.denhaag-columns):not(.denhaag-button-group) + .utrecht-heading-2`

paulAcato: @robbieatacato , in my opinion not. Since the columns, accordion and cta-link-group are all in the nl-design-system of  the Hague. If u apply it with *, i wish you luck with overwriting all these styles.

robbieatacato: @paulAcato I still cannot find these css classes with margins: 

- .accordion-group / .denhaag-accordion-group
- .cta-link-group / .denhaag-cta-link-group

If you mean:

- .denhaag-accordion, this component doesn't have spacing at the bottom (https://nl-design-system.github.io/denhaag/?path=/story/css-surfaces-accordion--default-story)
- .denhaag-cta-link, this component doesn't have spacing at the bottom (https://nl-design-system.github.io/denhaag/?path=/story/css-actions-cta-link--grouped)

paulAcato (approved):

PR #1133, opened 2022-10-18
Style/gdh 904 table

robbieatacato:
### Purpose:

- remove table row hover background color (because there is no interaction)

### Figma:

- https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Design-System---Gemeente-Den-Haag?node-id=5213%3A16235

closes

### DIFF
--- a/components/Table/src/index.scss
+++ b/components/Table/src/index.scss
@@ -99,7 +99,7 @@
 .denhaag-table__body > .denhaag-table__row:hover .denhaag-table__cell,
 .denhaag-table__body > .denhaag-table__row--hover .denhaag-table__heading,
 .denhaag-table__body > .denhaag-table__row:hover .denhaag-table__heading {
-  --denhaag-table-row-background-color: none;
+  --denhaag-table-row-background-color: transparent;
 }
 
 .denhaag-table__row--focus,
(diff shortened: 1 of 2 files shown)


components/Table/src/index.scss
```diff
@@ -99,7 +99,7 @@
 .denhaag-table__body > .denhaag-table__row:hover .denhaag-table__cell,
 .denhaag-table__body > .denhaag-table__row--hover .denhaag-table__heading,
 .denhaag-table__body > .denhaag-table__row:hover .denhaag-table__heading {
-  --denhaag-table-row-background-color: var(--denhaag-table-row-hover-background-color, transparent);
+  --denhaag-table-row-background-color: none;
```
paulAcato, line 102: Background-color: none is invalid.
Schermafbeelding 2022-10-18 om 14 21 44

paulAcato (changes requested):

components/Table/src/index.scss
robbieatacato: Gefixt, zie laatste commit.

paulAcato: Token still exists: https://github.com/nl-design-system/denhaag/blob/www.denhaag.nl/proprietary/Components/src/denhaag/table.tokens.json#L52

paulAcato (changes requested):
Token still exists, which is now redundant: https://github.com/nl-design-system/denhaag/blob/www.denhaag.nl/proprietary/Components/src/denhaag/table.tokens.json#L52

paulAcato (approved):

PR #1134, opened 2022-10-21
Fix/components keycloak login page

YourivHDenHaag:
Changes to fix some issues related to the Keycloak login page:

- Single button group is not full width on mobile
- CardAuthentication doesn't use the button group
- Images in story templates don't work in Storybook on Github
- Margins around body

And implement some design changes of the CardAuthentication component.

### DIFF
--- a/.storybook/stories/templates/Login-page.stories.mdx
+++ b/.storybook/stories/templates/Login-page.stories.mdx
@@ -29,7 +29,12 @@ export const navigationPathData = [
 ];
 
 <Canvas>
-  <Story name="Login page">
+  <Story
+    name="Login page"
+    parameters={{
+      chromatic: { viewports: [1280, 360] },
+    }}
+  >
     <Page header={<Header breadcrumb={<Breadcrumb navigationPath={navigationPathData} />} />}>
       <ResponsiveContent>
         <Heading2>Inloggen</Heading2>
--- a/.storybook/stories/templates/Login-page.stories.mdx
+++ b/.storybook/stories/templates/Login-page.stories.mdx
@@ -14,7 +14,12 @@ import clsx from "clsx";
 
 import "./story-login.css";
 
-<Meta title="Templates/Login page" />
+<Meta
+  title="Templates/Login page"
+  parameters={{
+    chromatic: { viewports: [1768, 1280, 768, 360] },
+  }}
+/>
 
 # Login page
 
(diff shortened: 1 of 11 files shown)


.storybook/stories/templates/Login-page.stories.mdx
```diff
@@ -29,7 +29,12 @@ export const navigationPathData = [
 ];
 
 <Canvas>
-  <Story name="Login page">
+  <Story
+    name="Login page"
+    parameters={{
```
Yolijn, line 34: Ah I would have thought these parameters to be part of the `<Meta>` so all login stories would be tested for various widths. Do you think it's enough to have this single story for mobile?

components/CardAuthentication/src/_mixin.scss
```diff
@@ -75,8 +86,19 @@
     margin-inline-end: auto;
     margin-block-end: var(--denhaag-card-authentication-portrait-logo-margin-block-end);
   }
+  .denhaag-card-authentication__title {
+    margin-block-end: 18px;
```
Yolijn, line 90: I think we might want to turn this into a design token. So `--denhaag-card-authentication-title-margin-block-end` and put the `18px` as a value in the design token file.

components/CardAuthentication/src/story-template.tsx
```diff
@@ -103,27 +103,29 @@ export const AuthenticationCard = ({
...
-              height="1em"
-              viewBox="0 0 24 24"
-              width="1em"
-              xmlns="http://www.w3.org/2000/svg"
-            >
-              <path d="M12.293 5.293a1 1 0 011.414 0l6 6a1 1 0 010 1.414l-6 6a1 1 0 01-1.414-1.414L16.586 13H5a1 1 0 110-2h11.586l-4.293-4.293a1 1 0 010-1.414z" />
-            </svg>
-          </span>
-        </a>
+        <div className="denhaag-button-group denhaag-button-group--single">
```
Yolijn, line 106: I think the `denhaag-button-group--single` doesn't exist anymore?

Yolijn (changes requested):

components/CardAuthentication/src/_mixin.scss
YourivHDenHaag: Done!

components/CardAuthentication/src/story-template.tsx
YourivHDenHaag: I've removed all `denhaag-button-group--single` and `denhaag-button-group--multiple` classes, since we don't use them anymore.

PR #1161, opened 2022-12-13
feat: add first iteration of footer component

YourivHDenHaag:
PR for https://github.com/nl-design-system/denhaag/issues/1143 (without mega navigation part).


.storybook/stories/templates/Login-page.stories.mdx
```diff
@@ -33,9 +32,161 @@ export const navigationPathData = [
...
+      ></path>
+    </svg>
+  );
+};
+
+export const footerSocialData = {
+  title: "Volg ons op social media",
+  links: [
+    { icon: <FacebookIcon />, label: "Facebook", url: "https://facebook.com" },
+    { icon: <TwitterIcon />, label: "Facebook", url: "https://facebook.com" },
```
Yolijn, line 153: These labels should be different for each 😜

components/Footer/src/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+import ResponsiveContent from '/components/ResponsiveContent';
+
+import './index.scss';
+
+export interface FooterListItemData {
+  label: string;
+  url: string;
+}
+
+export interface FooterLegalProps extends Omit<BaseProps, 'tabIndex' | 'classNamees' | 'children'> {
```
Yolijn, line 12: typo:`classNames`

Yolijn: Why do we want to Omit className? I think maybe we could use the className as one of the arguments of `clsx` instead. If somehow somewhere we then need a tiny addition or change we can just add an extra className to the `<FooterLegalProps className="fancy-footer-snowflake"/>` ?

components/Footer/src/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  copyrightLabel?: string;
+  socialData?: FooterSocialData;
+  newsletterData?: NewsLetterData;
+  contactData?: ContactData;
+}
+
+export const FooterLegal: React.FC<FooterLegalProps> = (props: FooterLegalProps) => {
+  return (
+    <section className="denhaag-footer__legal">
+      <div className="denhaag-responsive-content">
```
Yolijn, line 57: Not sure but maybe we could combine `denhaag-resonsive-content` with `denhaag-footer__legal` or `denhaag-footer-legal__container`?

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
+.denhaag-footer {
+  --utrecht-paragraph-color: #fff;
```
Yolijn, line 2: Hmmm I'm thinking maybe these should come from a token too?

`--utrecht-paragraph-color: var(--denhaag-footer-color)`

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+  --utrecht-paragraph-font-size: 16px;
+  --utrecht-paragraph-line-height: normal;
+  --utrecht-paragraph-margin-block-start: 12px;
+  --utrecht-paragraph-margin-block-end: 12px;
+
+  background-color: #4b4b4b;
```
Yolijn, line 8: And this background-color could then just be `background-color: var(--denhaag-footer-background-color)`

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+
+  --utrecht-link-social-background-color: transparent;
```
Yolijn, line 10: And these tokens could be set in the Proprietary folder too? Or do we use the `link-social` compponent on other (light) backgrounds too?

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+  --utrecht-list-social-item-margin-inline-end: 20px;
+  --utrecht-link-social-color: #fff;
+  --utrecht-link-social-hover-color: var(--denhaag-color-ocher-2);
+
+  .utrecht-list-social {
+    align-items: baseline;
+  }
+
+  .denhaag-footer__mega {
+    background-color: var(--denhaag-color-grey-5);
```
Yolijn, line 22: I would set the `denhaag.footer.mega.background-color` to `{denhaag.color.grey.5}` in JSON and get it here with `var(--denhaag-footer-mega-background-color)` 

For A11y tests later it might also be helpful to always set a text color too. So `denhaag-footer-mega-color` :)

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  .denhaag-list__item:nth-child(1) {
+    padding-inline-start: 0;
+  }
+
+  @media (max-width: 1279px) {
+    .denhaag-list__item {
+      padding-inline-start: 0;
```
Yolijn, line 65: 🤔

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  .denhaag-footer__follow-us-list {
```
Yolijn, line 69: Is the `denhaag-footer__follow-us-list` still used?

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+    justify-content: space-between;
+    list-style: none;
+    padding-inline-start: 0;
+
+    @media (min-width: 768px) {
+      justify-content: initial;
+    }
+  }
+
+  .denhaag-footer-group {
```
Yolijn, line 81: What is a `denhaag-footer-group`? I only see it used once but is it something we will use multiple times? Otherwise we might want to rename it to something like denhaag-footer__contact-container (not the best name, but something like it) :)

components/Footer/src/index.scss
```diff
@@ -0,0 +1,95 @@
...
+    display: grid;
+    gap: 32px;
+    padding-block-start: 40px;
```
Yolijn, line 84: Might be nice to get these hardcoded values from the JSON too?

Yolijn (approved):

PR #1186, opened 2023-01-31
Feature/sidenav+header components

YourivHDenHaag:


components/Breadcrumb/src/index.tsx
```diff
@@ -90,14 +59,9 @@ export const Breadcrumb: React.FC<BreadcrumbProps> = (props: BreadcrumbProps) =>
...
-    const rootClassNames = clsx(
-      'denhaag-breadcrumb__item',
-      shouldCollapseItem ? 'denhaag-breadcrumb__item--hidden' : '',
-    );
-
     return (
-      <li
-        className={rootClassNames}
+      <BreadcrumbListItem
+        collapse={shouldCollapseItem}
```
Yolijn, line 64: Maybe use `hidden` instead of `collapse`?

components/Breadcrumb/src/index.tsx
```diff
@@ -108,19 +72,26 @@ export const Breadcrumb: React.FC<BreadcrumbProps> = (props: BreadcrumbProps) =>
         ) : (
           <BreadcrumbItem item={item} isLastItem={isLastItem} contentNumber={contentNumber} />
         )}
-      </li>
+        <meta content={contentNumber.toString()} itemProp="position" />
```
Yolijn, line 75: Hmm I'm wondering if the `contentNumber` as it's content would already be enough? Or is it the typing requiring us to explicitly convert it to a string?

components/Breadcrumb/src/index.tsx
```diff
@@ -108,19 +72,26 @@ export const Breadcrumb: React.FC<BreadcrumbProps> = (props: BreadcrumbProps) =>
...
+      </BreadcrumbListItem>
     );
   });
 
   return (
-    <nav aria-label="Breadcrumb" className="denhaag-breadcrumb">
+    <BreadcrumbNavigation aria-label="Breadcrumb">
       <ResponsiveContent>
-        <ol className="denhaag-breadcrumb__list" itemScope itemType="https://schema.org/BreadcrumbList">
+        <BreadcrumbList itemScope itemType="https://schema.org/BreadcrumbList">
```
Yolijn, line 83: I think `itemScope={itemScope}` is needed here? now it's always going to be true

components/Header/src/Header.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { HTMLAttributes } from 'react';
+import clsx from 'clsx';
+import './index.scss';
+
+export type HeaderProps = HTMLAttributes<HTMLDivElement>;
+
+export const Header: React.FC<HeaderProps> = (props: HeaderProps) => {
```
Yolijn, line 7: I prefer destructuring the props instead of using `props.className`, do you prefer it the other way around?

components/Header/src/HeaderLogic.tsx
```diff
@@ -0,0 +1,247 @@
...
+  return (
+    <LinkGroup>
+      <Heading4>{props.label}</Heading4>
+      <LinkGroupList>{navigationLinks}</LinkGroupList>
+    </LinkGroup>
+  );
+};
+
+export const HeaderLogic: React.FC<HeaderLogicProps> = (props: HeaderLogicProps) => {
+  const menuConstants = {
```
Yolijn, line 66: Might be nice to get the breakpoint from the props instead? Maybe even with the 1024 as the default? 

`{mobileBreakpoint = 1024, ...props}: HeaderLogicProps`

components/Header/src/HeaderLogic.tsx
```diff
@@ -0,0 +1,247 @@
...
+                <Heading4>Kies uw voorkeurstaal</Heading4>
+                <LanguageSwitcherLogic {...props.languageSwitcherMenu}></LanguageSwitcherLogic>
+              </SheetContainer>
+            </ResponsiveContent>
+          </Sheet>
+          <SheetOverlay onClick={handleCloseSheet} />
+        </React.Fragment>
+      )}
+      {welcomeMenuActive && (
+        <React.Fragment>
```
Yolijn, line 194: **Side node**: In other projects I use `<></>` instead of `<React.Fragment>` maybe we can discuss which you prefer and change them in the codebase in a separate PR.

components/Header/src/index.scss
```diff
@@ -17,4 +18,24 @@
...
+
+  .denhaag-header__actions {
+    display: none;
+
+    @media (min-width: 1024px) {
+      display: flex;
+    }
+
+    .denhaag-header__actions-action {
+      margin-inline-start: 24px;
```
Yolijn, line 30: I think we might want to make this a CSS variable too?

components/LanguageSwitcher/src/index.scss
```diff
@@ -0,0 +1,61 @@
...
+    padding-inline-start: var(--denhaag-language-switcher-list-padding-inline-start);
+
+    @media (min-width: 1024px) {
+      padding-block-start: var(--denhaag-language-switcher-list-padding-block-start);
+      margin-block-end: var(--denhaag-language-switcher-list-margin-block-end);
+    }
+
+    .denhaag-language-switcher__list-item {
+      list-style: var(--denhaag-language-switcher-list-item-list-style);
+      padding-block-start: 16px;
```
Yolijn, line 19: Should these also be design tokens?

components/LanguageSwitcher/src/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+        >
+          {language.title}
+        </Link>
+      </LanguageSwitcherListItem>
+    );
+  });
+
+  return (
+    <LanguageSwitcher>
+      <LanguageSwitcherTitle className="utrecht-heading-4">{title}</LanguageSwitcherTitle>
```
Yolijn, line 54: I'm still not sure if adding a classname of a specific heading is a great idea or will cause headaches in the future as one would not expect this component to change necessarily when changing the properties of heading-4...
Okay for now, but something we might want to change later

components/Header/src/index.scss
YourivHDenHaag: Done!

components/LanguageSwitcher/src/index.scss
YourivHDenHaag: Done!

components/Breadcrumb/src/index.tsx
YourivHDenHaag: Yes, typing requires us to explicitly convert the contentNumber to a string.

components/Header/src/HeaderLogic.tsx
YourivHDenHaag: Done!

components/Header/src/HeaderLogic.tsx
YourivHDenHaag: I prefer `<></>`, i'm going to change this.

PR #1217, opened 2023-04-20
feat: added mobile menu, improved menu component

YourivHDenHaag:


components/Menu/src/MenuButton.tsx
```diff
@@ -123,7 +119,7 @@ export const MenuButtonExpandable: React.FC<MenuButtonProps> = ({
   );
 
   return (
-    <button id={props.id} onClick={props.onClick} className={rootClassNames} title={props.children?.toString()}>
+    <button className={rootClassNames} title={props.children?.toString()} {...props}>
```
Yolijn, line 122: Hmm why do we need a title here? We won't know for sure the children are never going to be a string only do we? It might include a `strong` or `icon` or something?

Yolijn: I do see it was already there, but I would choose to just use `{...props}` and let the layers above decide if one of the props is `title`

YourivHDenHaag: You're right, i've changed this.

Yolijn (approved):

PR #1299, opened 2023-08-10
fix: description-list caption class

### LINKED ISSUE #1296: Design-token voor de data-list title
- [ ] Voeg een design-token toe voor de (CSS) data-list title
- [ ] Zet deze token in het Den Haag thema op de juiste waarde 
https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Den-Haag-Design-System?type=design&node-id=8487-27946&mode=design&t=ZeR6o0deD3elPkIy-4

Borai:
Fixed incorrect classname 

Closes #1296

### DIFF
--- a/components/DescriptionList/src/index.scss
+++ b/components/DescriptionList/src/index.scss
@@ -36,6 +36,7 @@
 }
 
 .denhaag-description-list__title {
+  color: var(--denhaag-description-list-title-color);
   font-weight: var(--denhaag-description-list-title-font-weight, bold);
   padding-block-end: var(--denhaag-description-list-title-padding-block-end, var(--denhaag-description-list-padding));
   padding-block-start: var(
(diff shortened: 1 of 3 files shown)


Yolijn (changes requested):
If this closes the issue about adding a design token this PR is not yet complete. For the description-list there's a `caption` and a `title`. The `caption` is the optional heading above the description-list. The title is the `dt` which occurs for every row in the list. 
According to the Figma design the title should have a different color than the detail. However neither has a `color` property. 

This issue was about adding the `color` property to the `title` in this file https://github.com/nl-design-system/denhaag/blob/main/components/DescriptionList/src/index.scss with a reference to a new design token in this file https://github.com/nl-design-system/denhaag/blob/main/proprietary/Components/src/denhaag/description-list.tokens.json

Borai:
@Yolijn Sorry I was confused about the naming in the ticket. I commit some new code.

Yolijn (approved):
No worries, this is adding the right color 😊

PR #1302, opened 2023-08-17
Feature/tab component refactor

YourivHDenHaag:
Refactored the Tab component and removed the dependency on Material UI.

Disabled tabs and a mobile overflow wrapper will be added later.


markm-ictu:
Looks good, as far as I can see, I don't see anything notable 👍

components/Tab/src/Tab.tsx
```diff
@@ -1,63 +1,25 @@
...
-      disableRipple
-    />
+    <div {...props} ref={ref} role="tab" aria-selected={selected} className={rootClassNames}>
+      {props.children}
+    </div>
   );
-};
+});
+
+Tab.displayName = 'Tab';
```
Borai, line 23: Do we need this extra displayName if we are using a const?

YourivHDenHaag: @Borai It's necessary to prevent this eslint error:
`denhaag\components\Tab\src\Tab.tsx
  14:20  error  Component definition is missing display name  react/display-name`

Yolijn (approved):

PR #1309, opened 2023-08-24
1305 react + css file component

Borai:
WIP: Please let me know if I'm on the right track here.

- [ ] Utrecht-link the right component?
- [ ] use svg icon the right way?


components/File/README.md
```diff
@@ -21,7 +21,7 @@ The file download consists of:
 
 ## (Interactive) states
 
-The file download contains the states normal, hover and focus.
+The file contains the states normal, hover and focus.
 
 ## Design properties
```
YourivHDenHaag, line 27: U can use the readme from here: https://www.figma.com/file/JpoY3waVoQGlLQzQXTL9nn/Den-Haag-Design-System?type=design&node-id=5418-16936&mode=dev

components/File/src/index.scss
```diff
@@ -0,0 +1,56 @@
+.denhaag-file {
+  display: flex;
+  color: inherit;
+  text-decoration: inherit;
+  border: var(--denhaag-file-border);
+
+  .utrecht-link {
```
YourivHDenHaag, line 7: I think we don't want to override CSS properties like this, but we want to override the design tokens of the Utrecht Link if possible. (https://github.com/nl-design-system/utrecht/blob/main/components/link/css/_mixin.scss). @Yolijn What's your opinion about this?

components/File/src/stories/bem.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+          />
+          <path d="M13 2.5V9.5H20" stroke="#4B4B4B" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" />
+        </svg>
+      </div>
+      <div class="denhaag-file__right">
+        <div class="denhaag-file__label">
+          <span id="name">File name </span>
+          <span id="description">(pdf, 2MB, 1-1-2023)</span>
+        </div>
+        <div class="utrecht-link utrecht-link--icon-left" tabindex="-1">
```
YourivHDenHaag, line 96: I think we want the icon as svg element inside the download link, like in the Den Haag link: https://nl-design-system.github.io/denhaag/?path=/docs/css-navigation-link--link-with-icon-before#link-with-icon-before-hovered, because I think is somewhat more flexible and in the React component we could use the <DownloadIcon /> component from the @gemeente-denhaag/icons package. But i see Utrecht does this with an icon in the CSS, @Yolijn what's your opinion about this?

components/File/src/index.scss
Yolijn: Yes I agree. We want to try not to override properties but design tokens.

Also we want to prevent Den Haag brand in the component. So no `--denhaag-space-xl` but `--denhaag-file-padding-inline-start` for example

In this case I would assume something like this

```css
.denhaag-file__link {
 --utrecht-link-padding-inline-start: var(--denhaag-file-link-padding-inline-start);
 background-position: ...
 background-image: ...
}

.denhaag-file__link:hover {
 --utrecht-link-color: var(--denhaag-file-hover-color, --utrecht-link-hover-color);
}
```

components/File/src/index.scss
```diff
@@ -0,0 +1,59 @@
+.denhaag-file {
+  display: flex;
+  color: inherit;
+  text-decoration: inherit;
+  border: var(--denhaag-file-border);
```
Yolijn, line 5: To make sure our design tokens will also work in other platforms later we unfortunately cannot use these CSS specific shorthand properties. So for NL Design System we propose to split them up into logical properties: `border-width` `border-color` `border-style` (which might be a hardcoded `solid` if dashed and others won't make sense. 
b

components/File/src/index.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  &--focus,
+  &:focus {
+    outline: var(--denhaag-file-focus-outline);
```
Yolijn, line 9: For outline we also don't use the shorthand but the separate constituent properties: https://developer.mozilla.org/en-US/docs/Web/CSS/outline#constituent_properties

proprietary/Components/src/denhaag/file.tokens.json
```diff
@@ -0,0 +1,32 @@
+{
+  "denhaag": {
+    "file": {
+      "border": { "value": "1px solid {denhaag.color.grey.2}" },
+      "hover": {
+        "color": { "value": "{denhaag.color.blue.4}" },
+        "background-image": {
```
Yolijn, line 7: Is this background-image still in use? I don't see it in the css, or did I overlook it?

PR #1321, opened 2023-09-19
fix: change in design related to focus/hover state + cleanup some code

YourivHDenHaag:
Tab component:
- Fixed missing uuid dependency
- Applied changes in design to component
- Removed duplicate Tabs export by removing unnecessary Tabs.tsx


components/Tab/package.json
```diff
@@ -28,6 +28,10 @@
     "build": "rollup -c ../../rollup.config.js",
     "clean": "rimraf dist tsconfig.tsbuildinfo"
   },
+  "dependencies": {
+    "@types/uuid": "9.0.2",
```
Borai, line 32: Can we move this to dev devs?

YourivHDenHaag: Yes, good point, i'll fix this.

YourivHDenHaag: Done

Borai (approved):

PR #1329, opened 2023-09-26
chore: fix lint errors that will occur with new major versions

Robbert:
We should merge these ahead of the upgrade PR, since we don't know how long it will take to finish that branch.


Borai (approved):
LGTM 🚀

PR #1330, opened 2023-09-26
chore: fix lint errors that will occur with new major versions (partial)

Robbert:


components/Accordion/src/index.tsx
Borai, line 1: Do we need to fill those index file with exports of the particular component?

Robbert: Yes, that is what is happening everywhere. The readme and one package.json (probably copied from the readme) was outdated, all other package.jsons were okay.

Borai: So I this something we need to create a task for?

Robbert: No, this was the last one.

PR #1344, opened 2023-09-27
iso date

Borai:
Use iso date in action component


YourivHDenHaag (approved):
Fine for now due to time pressure, but we will have to improve this component later (think of hardcoded texts and generic solution for datetime).

Perhaps we should also add here that this is an experimental component, which will not be fully backwards compatible at this moment

Yolijn:
Stuk beter zo! Is altijd goed opletten met die storybook in Chromatic zodat we niet elke dag changes hebben.

PR #1524, opened 2024-04-15
feat(hero): modified hero link

marcoderover:
Added font-weight token for the link in the hero
Added font-weight style for the link in the hero


paulAcato (approved):
Approved, `.denhaag-link` doesn't have any font-weight token. If it had a font-weight token you could have overwritten it in the `.denhaag-hero--theme-image {}`

PR #1526, opened 2024-04-25
docs: add sidenavigation docs to story

bddjong:
The docs for the sidenav were still using button docs. Set them to the correct docs for the CSS and React story


YourivHDenHaag (approved):
Thanks!

PR #1571, opened 2024-10-16
fix: correct margin between label and badge

YourivHDenHaag:


components/Sidenav/src/index.scss
```diff
@@ -40,6 +40,11 @@
   padding-block-end: var(--denhaag-sidenav-link-padding-block-end);
 }
 
+.denhaag-sidenav__link-label {
+  display: inline-flex;
```
Borai: Deze

Borai (approved):

PR #1706, opened 2025-03-06
Bugfixes www-denhaag-nl/gdhs-60

paulAcato:
- Set marge between paragraph and image-banner, set to "Friendly" due to input of design.
- Remove offset of 10px in the steps, the 10px was causing a whitespace;
- Added a smaller preview, due to the issue only appear on smaller screens.

Cleanup closed branches:
Closes #1675 #1620 #1698 #1697

### DIFF
--- a/components/ProcessSteps/src/Step.scss
+++ b/components/ProcessSteps/src/Step.scss
@@ -1,7 +1,6 @@
 .denhaag-process-steps__step {
   position: relative;
-  padding-block-end: 0;
-  padding-block-start: 0;
+  padding-block: 0;
 }
 
 .denhaag-process-steps__step + .denhaag-process-steps__step {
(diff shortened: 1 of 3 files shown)


merelacato (approved):

components/ProcessSteps/src/Step.scss
```diff
@@ -13,8 +12,8 @@
   content: "";
   position: absolute;
   left: calc(max(var(--denhaag-process-steps-step-marker-size, var(--denhaag-space-lg)), var(--denhaag-space-2xl)) / 2);
-  top: calc(-1 * (var(--denhaag-process-steps-step-distance) + 10px));
```
rmpel, line 16: This change makes the problem bigger.
please use a -20px for top and a +20px for height. (in other words;

```
top: calc(-1 * var(--denhaag-process-steps-step-distance) + 20px);
height: calc(var(--denhaag-process-steps-step-distance) + 20px);
```

PR #1788, opened 2025-06-11
Feature/gdh 1690 style(link-grid): remove margin and add border to link-grid without image

FrankaBeekman:
Removed inline margin for link-grid without image, and fixed broken border on the left side of link-grid without image.


paulAcato (changes requested):
I think we should implement denhaag-container in here?

Setting the margin-inline to 0 will cause more troubles i think?

Maybe check on test.denhaag.nl with replacing the styling with this in all cases.

paulAcato (approved):

PR #1866, opened 2025-09-04
GDH-1524 style(procedure): add and fix margins for procedure

FrankaBeekman:
add more specificity so that it actually overwrites the <ol> and added top margin when under paragraph

### DIFF
--- a/components/Procedure/src/index.scss
+++ b/components/Procedure/src/index.scss
@@ -25,7 +25,8 @@
 }
 
 .denhaag-procedure-container {
-  [class*="utrecht-heading-"] + & {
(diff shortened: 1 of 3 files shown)


components/Procedure/src/index.scss
```diff
@@ -25,7 +25,8 @@
 }
 
 .denhaag-procedure-container {
-  [class*="utrecht-heading-"] + & {
+  [class*="utrecht-heading-"] + &,
```
paulAcato, line 28: Why not `[class*="utrecht-"]` ?

FrankaBeekman: Would that only impact headings and paragraphs?

PR #1887, opened 2025-10-02
style(image): add variable to use in other components that have a den…

FrankaBeekman:


paulAcato:
Closed, as discussed.

PR #1916, opened 2025-11-06
GDH-1729 style(accordion): modernize accordion and fix animation

FrankaBeekman:
convert accordion to a details-summary setup
dynamically add inert and aria-expanded to summary and inner div for accessibility
change lint rules to allow ::details-content pseudo element (currently supported by all main browsers for multiple versions)

### DIFF
--- a/components/Accordion/src/index.scss
+++ b/components/Accordion/src/index.scss
@@ -94,7 +94,7 @@
   outline: 0;
 }
 
-.denhaag-accordion__container[open] .denhaag-accordion__panel .denhaag-icon {
+.denhaag-accordion__container[open] .denhaag-accordion__panel > .denhaag-icon {
   --denhaag-accordion-icon-transform: rotate(180deg);
 
   color: var(--denhaag-color-green-3);


components/Accordion/src/AccordionToggle.js
```diff
@@ -1,59 +1,25 @@
...
-      // Set opposite aria-expanded.
-      element.setAttribute('aria-expanded', expanded);
-    };
-  });
-};
-
-export default AccordionToggle;
+export default class Accordion {
+  constructor(className = 'denhaag-accordion__container') {
+    this.collapses = document.getElementsByClassName(className);
```
paulAcato: constructor is called before DOMContentLoaded -> so, there's a possibility getElementsByClassName returns none.

Might want to add an `initialize()` method?

components/Accordion/src/AccordionToggle.js
```diff
@@ -1,59 +1,25 @@
...
+  }
+
+  events() {
```
paulAcato: Why not like?

```
events() {
  if( ! this.collapses  || !this.collapses.length) {
     return;
  }

    Array.from(this.collapses)?.forEach( collapse => collapse?.addEventListener('toggle', () => this.toggleAttributes(collapse) ) );
  }

File: components/Accordion/src/index.scss [hunk 3]
```diff
@@ -68,61 +76,47 @@
...
-.denhaag-accordion__title--disabled::before,
-.denhaag-accordion__title[disabled]::before {
-  content: "";
-  width: 100%;
-  height: 100%;
-  position: absolute;
-  top: 0;
-  left: 0;
+.denhaag-accordion__container--disabled,
+.denhaag-accordion__container[inert] {
```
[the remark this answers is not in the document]
FrankaBeekman: disabled doesn't work on details/summary

components/Accordion/src/AccordionToggle.js
```diff
@@ -1,59 +1,25 @@
...
-      }
-
-      // Set opposite aria-expanded.
-      element.setAttribute('aria-expanded', expanded);
-    };
-  });
-};
-
-export default AccordionToggle;
+export default class Accordion {
```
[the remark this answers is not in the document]
paulAcato: discussed, no IDS!

components/Accordion/src/AccordionToggle.js
paulAcato: Discussed, can be fired on constructor since class is that tiny

PR #2047, opened 2026-05-07
feat: replace file wrapper with div and add anchor and real button

### LINKED ISSUE #2026: Feat: improve File component for accessibility and add more styling options
### User story

Open Inwoner (OIP) is using the Den Haag 'File' component. This component could be improved for accessibility and is missing some styling/values. An overview with suggested improvements (as already implemented by OIP itself) can be found in the NLDS discussions here: https://github.com/orgs/nl-design-system/discussions/347#discussioncomment-16208598
This feature request collects the different findings from there.

### Describe the solution you'd like

Here is a collection of our findings so that you can choose any of these, or even all of them if this agrees with accessibility research:

#### design-tokens:

- ~~add a border-radius token, we use rounded borders but there is no design-token for `.denhaag-file -> border-radius`~~
- add tokens to style the extension text

#### HTML/structure

- not making the entire File component a link, since this could be bad for accessibility/screenreaders, OIP made the whole element into a Div, so that only the 'button' like element inside will have its semantic behaviour: https://github.com/maykinmedia/open-inwoner/blob/88630a45eeaeca5fd4ecbe0b074cc9cb226ffb8f/src/open_inwoner/react/components/File/File.tsx#L92 also viewable in OIP Storybook: https://maykinmedia.github.io/open-inwoner/?path=/docs/components-file--docs
- adding some kind of slot so we can use our own Icons (not just SVG)
- OIP currently uses its own logic to either show an 'image/painting' like icon if the file is an image type, or shows a 'document' like icon if it is a type of text file. It is not documented yet in NLDS whether this is a required feature or not.
- add a variant where there is no Download button and no Remove functionality either, since we have a couple of 'preview' situations where no functionality is needed: https://maykinmedia.github.io/open-inwoner/?path=/story/components-file--without-delete-or-link
- add extra spans so that the File extension, and upload-dates can be styled. 
- it's better to add javascript like the remove-function if that element is a true HTML Button

#### CSS/styling

- since the whole File component is a Div here, we needed styling for the Download `<a>` link, like `text-decoration: none;` which perhaps could be design-tokens
- set the hard-coded cursor as default if the whole element is no longer a link
- same text-decoration styling was needed for the real 'delete/remove' button, (wither border styling needed to be removed from the Button), but it also needs more styling/design-tokens-as-a-button: https://github.com/maykinmedia/open-inwoner/blob/88630a45eeaeca5fd4ecbe0b074cc9cb226ffb8f/src/open_inwoner/react/components/File/File.scss#L28
- OIP has always displayed the File extensions in all-caps, with its own `text-transform: uppercase;` style which should perhaps be left outside of the design-tokens.

### Describe alternatives you've considered

We are only using the tokens and CSS from Den Haag (so not the React component), and have rebuilt the structure as a Preact/web-component Div ourselves, but still using the same class-names and design-tokens.
If we needed more structure or styling, we added the `oip-` prefix to these elements.

### Additional context

OIP added the Den Haag File component to NLDS Themes repo here: https://github.com/nl-design-system/themes/pull/1353
and added an Utrecht status badge here: https://github.com/nl-design-system/themes/pull/1381

UPDATE: new feedback from accessibility officer April 9th 2026:

> "...Voor een screanreadergebruiker is het vooral belangrijk dat alle context aanwezig is op het interactactieve element. Of dan het klikbare gebied vergroot is, of het hele component klikbaar is, maakt an sich niet zo veel uit.
> 
> In die van Den Haag mist nu de informatie 'Download' en wordt alleen de bestandsnaam meegegeven. Daarnaast is de tekst Download nog wel bereikbaar met de pijltoetsen (maar niet via de pijltoetsen met een screenreader), dus daar gaat ook iets mis...." 

### Acceptance criteria

1. When we implement the File component in our project, it needs to have the same border-radius as all our other elements, 
2. The 'delete button' behaves like a true button so that screenreaders can detect it as separate from the whole File element and give feedback on its javascript functionality.

jiromaykin:
Replacing the wrapper with a div element so that the entire File is only visually a link (so is still hoverable completely with the `::after` style), but in code will contain seprate elements that conform to accessibility.
Also: using a true Button for when the File component is used to Remove a file.

in this PR:

- [x] Remove the `<Wrapper>` pattern entirely
- [x] Change the root to a `<div>`
- [x] Make the download a real `<a>` tag
- [x] Make the delete/removable a real `<button>` tag
- [x] Remove the `tabIndex={-1}` from the inner divs
- [ ] check/improve accessibility issues
- [x] make a CSS :after to wrap the 'Download' version of file component so that it looks like a very large link.

still to do and still in Open Inwoner:

- oip- prefixed classes that add things Den Haag simply doesn't have, like the extension/size span styling and text-transform: uppercase
- MaterialIcon instead of SVG icons
- OIP formatFileSize logic
- OIP deleteUrl + confirm() handler
- OIP isImage prop and image icon logic so icon will look different if type of file is text or image or PDF... etc.
- lastUpdated formatting differences

**Closing issues**

Closes #2026

### DIFF
--- a/.changeset/sour-frogs-ring.md
+++ b/.changeset/sour-frogs-ring.md
@@ -1,0 +1,7 @@
+---
+"@gemeente-denhaag/storybook": minor
+"@gemeente-denhaag/design-tokens": minor
+"@gemeente-denhaag/file": minor
+---
+
+replace file wrapper with anchor and add real button
--- a/components/File/src/File.tsx
+++ b/components/File/src/File.tsx
@@ -17,7 +17,7 @@ interface FileProps {
 
 export const File = ({
   name,
-  href = '#',
+  href,
   size,
   lastUpdated,
   className,
@@ -26,7 +26,7 @@ export const File = ({
   removableLabel = 'Verwijderen',
   onClick,
 }: FileProps) => {
-  const extension = href?.lastIndexOf('.') >= 0 ? href.substring(href.lastIndexOf('.') + 1, href.length) : undefined;
+  const extension = href && href.lastIndexOf('.') >= 0 ? href.substring(href.lastIndexOf('.') + 1) : undefined;
   const lastUpdatedDate = lastUpdated ? new Date(lastUpdated).toLocaleDateString() : null;
   const FileTypeIcon = ({ ...props }) => {
     switch (extension) {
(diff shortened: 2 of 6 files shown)


jiromaykin:
Closing this issue because:
 for now we want to preserve the difference between Action-single (completely clickable) and Action-multiple (only the button should be clickable)

